PR #16, opened 2018-12-17
do not test on windows x86

heliosdrm:


Datseris:
I wonder where this problem was coming from... :S Something with accessing Sparse matrices with `Int32`, it definitely comes from Base...

Let's hope this doesn't come back to haunt us :D :D :D

heliosdrm:
It seems so: the error was that there is code trying to access regions of the sparse matrices using indices of type `Int64`, but the available methods in a winx86 build only accept `Int32`. The question is why the indices for accessing are signed as `Int64` in a 32-bit system.

In the code of `windowed.jl` there are vectors or indices signed as `Int`, but in 32-bit systems this should be `Int32`, according to https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/#Integers-1

heliosdrm:
No, I was wrong. Seems fixed by #17

Issue #15, opened 2018-12-17
What's wrong with Windows x86?


heliosdrm:
Appveyor fails for the x86 versions. No idea of what's happening there.

Datseris:
Yeah, there is a problem with the sparce matrices... You can see the error log here: https://ci.appveyor.com/project/JuliaDynamics/recurrenceanalysis-jl/branch/master/job/ny4vad40m5ks96yk#L74 Seems like this is a Base problem...?

Datseris:
Ah its on the `gmi` tests anyway. I'll remove them since we are moving this functionality to DelayEmbeddings.

PR #17, opened 2018-12-17
try test on win32 again

heliosdrm:
Found that there was some code in `matrices.jl` explicitly creating the recurrence matrix with `Int64` indices. Changed to see if win32 works.


heliosdrm:
Fix #15

Issue #21, opened 2018-12-26
Dedicated types for Recurrence matrices


Datseris:
as discussed in #20 having dedicated types RecurrenceMatrix, CrossRecurrenceMatrix etc, all of which are subtypes of AbstractRecurrenceMatrix has a lot of benefits. Let's discuss the implementation here.

1. What _type_ parameters should be there, besides the ones of the data field?
2. Any other fields besides the data field?

---

To-dos:

- [ ] Create types
- [ ] Ensure `recurrencematrix` and co return the types
- [ ] Enforce that the user is passing a recurrence matrix to e.g. laminarity, because the dispatch happens on ::AbstractRecurrenceMatrix (so far there is no guarantee that x is not a unfitting matrix)
- [ ] Pretty printing (which I super love). So when the user sees x they will get info that it is a recurrence matrix, etc. This is really cool for front end users!

Datseris:
I am opening a branch for this. At the moment I do not see any reason for recurrence and cross recurrence matrices to be of different type. I will make a datastructure that is simply called `RecurrenceMatrix` and has one field the data and the other field the type, simply as a string: `cross-`,`joint-` or the empty string.

heliosdrm:
I see your point about the recipe. Then, I think that the best is to keep the original approach: the types will be `RecurrenceMatrix`, etc. (I would not call `xxxPlot` to something that is not directly plottable).

Then, the question is about the name of the function to create the plottable matrix. Shall we keep `recurrenceplot`, or change it to `plottable`?

I'm opening a new issue about the other question (different types for recurrence and cross-recurrence matrices). This is related to your latest question in #20.

Datseris:
Closed in #25

Issue #20, opened 2018-12-23
1.0.0 release


Datseris:
For stability reasons, a 1.0 release is necessary, so that we can start following SymVer as far as breaking changes are concerned. @heliosdrm is there anything missing or anything you would like to do before we do this?

heliosdrm:
One question to decide on is what should be the default metric. The literature often advocates for the maximum norm, which is the simplest, and I followed that criterion. But it looks that the Euclidean norm is in this case more efficient because of BLAS optimization. Comparing with other software does not give a definitive answer: pyunicorn (Python) uses the maximum norm as default, whereas crqa (R) only use the Euclidean norm.

Besides, there are other features (not that much "breaking", but rather "extending" the functionality of the packages), which @pucicu suggested to me privately some time ago:
* export diagonal histograms.
* calculate some parameters used in cross-recurrence network analysis (as pyunicorn does).

Besides, I have thought on adding RQA parameters related to whitespace vertical structures. And more ambitious - for the long term, allow paralellization - calculation of the recurrence plots and RQA parameters is quite fit for that feature.

Perhaps we could categorize those features as milestones for 1.0, and following minor/major versions? 
This is extending functionality

heliosdrm:
I don't think that conversion to `Matrix` will happen so often: only if the embedding dimension is equal or greater than 10, which is not usual.

I'll create the individual issues tomorrow or the day after.

Datseris:
> I don't think that conversion to `Matrix` will happen so often: only if the embedding dimension is equal or greater than 10, which is not usual.

Yeah, forgot about that, you are right! But coming back to the performance, I checked again the first post on https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/pull/19 and I got reminded that the speed difference for large `N` is pretty huge: `Euclidean()` totally wins over `Chebyshev()` and also `Citiblock`. In fact, `Chebyshev` is by far the slowest one of the three, which is a very strong argument for changing the default norm :D

heliosdrm:
Yet another thing that I have in mind since the beginning, but I have not yet decided on it:

[...]

PR #25, opened 2018-12-31
[WIP] Dedicated structs for recurrence matrices

Datseris:
See discussions in #21 , #20

### DIFF
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -28,15 +28,6 @@ export embed,
        sorteddistances,
        @windowed
 
-# column values in sparse matrix (parallel to rowvals)
-function colvals(x::SparseMatrixCSC)
-    cv = zeros(Int,nnz(x))
-    @inbounds for c=1:size(x,2)
-        cv[nzrange(x,c)] .= c
-    end
-    cv
-end
-
 include("matrices.jl")
 include("plot.jl")
 include("rqa.jl")
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -32,6 +32,6 @@ include("matrices.jl")
 include("plot.jl")
 include("rqa.jl")
 include("radius.jl")
-include("windowed.jl")
+# include("windowed.jl")
 
 end
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -5,6 +5,7 @@ import Base.Meta.parse
 
 export embed,
        reconstruct,
+       Dataset,
        distancematrix,
        recurrencematrix,
        crossrecurrencematrix,
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -3,6 +3,7 @@ module RecurrenceAnalysis
 using Distances, Statistics, LinearAlgebra, SparseArrays, DelayEmbeddings, StaticArrays
 import Base.Meta.parse
 
+export RecurrenceMatrix, CrossRecurrenceMatrix, JointRecurrenceMatrix
 export embed,
        reconstruct,
        Dataset,
@@ -22,9 +23,6 @@ export embed,
        trappingtime,
        maxvert,
        rqa,
-       autocorrelation,
-       ami,
-       gmi,
        sorteddistances,
        @windowed
 
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -32,6 +32,6 @@ include("matrices.jl")
 include("plot.jl")
 include("rqa.jl")
 include("radius.jl")
-# include("windowed.jl")
+include("windowed.jl")
 
 end
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -89,6 +89,7 @@ end
 # Type
 #######################
 abstract type AbstractRecurrenceMatrix end
+const ARM = AbstractRecurrenceMatrix
 struct RecurrenceMatrix <: AbstractRecurrenceMatrix
     m::SparseMatrixCSC{Bool,Int64}
 end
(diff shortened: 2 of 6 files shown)


Datseris:
At the moment it looks like this:
```
julia> jointrecurrencematrix(x,y,0.1)
Joint recurrence matrix of size (1000, 1000) with 37462 entries:
  [1   ,    1]  =  true
  [2   ,    1]  =  true
  [3   ,    1]  =  true
  [19  ,    1]  =  true
  [24  ,    1]  =  true
  [33  ,    1]  =  true
  [64  ,    1]  =  true
  [66  ,    1]  =  true
  [72  ,    1]  =  true
  ⋮
  [707 , 1000]  =  true
  [708 , 1000]  =  true
  [735 , 1000]  =  true
  [789 , 1000]  =  true
  [856 , 1000]  =  true
  [927 , 1000]  =  true
  [985 , 1000]  =  true
  [999 , 1000]  =  true
  [1000, 1000]  =  true
```

Datseris:
@heliosdrm I will have to update all the `recurrencerate` and co. functions. I've noticed something weird in the source: e.g. in `recurrencerate`:
```
    typeof(0.0)( (count(!iszero, x)-theiler_nz)/prod(size(x)) )
```
first of all, why `typeof(0.0)` and not `Float64` ? (the latter is also not necessary as `/` always returns `Float64`)

secondly, why `count(!iszero, x)` ? the function `nnz(x)` gives the amount of nonzero elements.

Can I do those substitutions already? Also, `prod(size(x))` is equivalent with `length(x)` for our matrices.

Datseris:
Hm, @heliosdrm I **must** have messed up somewhere. I've reformed the core source code (no plotting / windowed yet) to use the new types. But, from the tests, these ones:
```
@test rqapar["DET"] == 364/426
@test rqapar["L"] == 364/50
@test rqapar["ENT"] ≈ 2.02058292
```
are now failing. They have close values but not exact anymore. Do you see the obvious mistake? If not I'll start searching again later. What confused me is that there are two methods for `verticalhistogram, diagonalhistogram`: one for sparse matrices one for normal matrices. I may have messed up there...

heliosdrm:
I'll check...

heliosdrm:
Ok, I found It. A bug in the method `verticalhistogram` that was not being hit before, but it is now. I'll fix it tomorrow.

heliosdrm:
Happy new year! Fixed the issue with the histograms.
I have also started to fix the windowed methods, but they are still failing because `iterate` is not yet defined for the new types.

Datseris:
Happy new year @heliosdrm !!! Thanks for the fix!  I am extending `iterate` now then :)

Datseris:
Extending iterate was so trivial, oh man I love metaprogramming. Windowed now works with `Recurrence` and `CrossRecurrence`, but not with `JointRecurrence`. The problem happens here:
```
            ret_ex = quote
                $ex_rmx
                $ex_rmy
                JointRecurrenceMatrix($rm1 .& $rm2)
            end
```
where `$rm1 .& $rm2` produces a `BitArray` instead of a `SparseArray`. Do you want me to extend `JointRecurrenceMatrix` to be able to accept `BitArray`s? I don't know how much this will mess up with conciseness of the code, so if we can avoid it, it may be better.

heliosdrm:
I think It would be better to make `rm1 .& rm2` be `SparseMatrixCSC` before passing it to `JointRecurrenceMatrix`

Datseris:
I don't know how to do that... :S Due to metaprograming it tells me that SparseArrays is not defined... :S

heliosdrm:
But `SparseArrays` is imported by `RecurrenceAnalysis`  so `RecurrenceAnalysis.SparseArrays`  should be available.

Datseris:
I've named the field of the matrices `.m` so far. Should we consider making it `.data` instead, to be mroe intuitive with respect to the `Dataset` ? The underlying data structure is of course different, but maybe consistency will make memorization easier.

Datseris:
re: on the SparseMatrix on `@window` : `RecurrenceAnalysis.SparseMatrixCSC` works.

But I still don't know what to do as I don't know what is the `Bitarray` is and how I should convert it to a sparse matrix...

heliosdrm:
Try this:
 ```
       ret_ex = quote
             $ex_rmx
             $ex_rmy
             JointRecurrenceMatrix($rm1.m .& $rm2.m)
       end
 ```

This should give `JointRecurrenceMatrix` an argument that is a sparse array, instead of a bit array.

Datseris:
Alright, that fixed it. There is now only 2 left failing  : 
```
@windowed(rrw = recurrencerate(rmatw), width=50, step=40)
@windowed rqaw = rqa(rmatw) width=50 step=40
```
these two lines give method missmatch. They try to pass a `SparseMatrix` into `recurrencerate` and `diagonalhistogram`. I'll leave this to you only because the `@windowed` code is too complex for me to understand :D

heliosdrm:
Ok, the windowed methods seem fixed (and I have also cleaned up a bit the code).

A remaining question is: now that RQA functions only dispatch on `AbstractRecurrenceMatrix`, is it still necessary to call the function for RQA entropy `rqaentropy`? There is no risk of confusion with other `entropy` methods anymore.

Datseris:
I'd still vote to keep it `rqaentropy`. By itself `entropy` is vague, because there are many entropies, and each one can be computed for many quantities. 

The function however is very specific: computes the Shannon entropy of the diagonal lengths, so I'd say `rqaentropy` is better.

pucicu:
I would like to add here that there are different entropy measures derived from the recurrence matrix. It might be a good idea to keep this in mind when defining the names for entropies here. Some variants are:

– entropy of the (black) diagonal lines (the usually used RQA measure “entropy”)
– entropy of the (black) vertical lines (not yet published but mentioned in G. Leonardi: A Method for the computation of entropy in the Recurrence Quantification Analysis of categorical time series, Physica A, 512, 824–836p. (2018)); extendable to the area distribution of block structures in RPs (see mentioned paper)
– entropy of the white vertical lines (= recurrence times) which is also known as “recurrence time entropy RTE” or “recurrence period density entropy RPDE”, (M. A. Little, P. E. McSharry, S. J. Roberts, D. A. E. Costello, I. M. Moroz: Exploiting Nonlinear Recurrence and Fractal Scaling Properties for Voice Disorder Detection, BioMedical Engineering OnLine, 6(23), 1–19p. (2007))
– different possibilities to estimate K2, e.g. using the Grassberger-Procaccia algorithm, or the approach by Faure et al(P. Faure, H. Korn: A new method to estimate the Kolmogorov entropy from recurrence plots: its application to neuronal signals, Physica D, 122(1–4), 265–279p. (1998))

We are currently working on a review about these measures which will also suggest some important fine tunings of the “RQA entropy”. I will let you know as soon as our ideas can be published.

Best wishes
Norbert

Datseris:
Thanks Norbert for all this useful information!

I am wondering whether this function is necessary at all. In the package `ChaosTools` the function `genentropy` computes the generalized (Renyi) entropy, which also includes the Shannon entropy. 

On the other hand, I can see that it is useful to have an entropy estimate to be included in the `rqa` bundle function. And `ChaosTools` is certainly too heavy of a dependency to add it here.

On the _other_ other hand, as Norbert mentioned, you can choose many different aspects of the `RecurrenceMatrix` and compute their entropy. After the aforementioned review paper is out, maybe the different entropies are too many and we can safely drop `rqaentropy` in favor of `genentropy` and simply calling `diagonalhist` or `verticalhist`. We'll see as this progresses I guess.

Issue #24, opened 2018-12-31
Calculation of RQA parameters for "white vertical structures"


heliosdrm:
Create the dedicated functions, and extend `rqa` to support the new related keys.

heliosdrm:
@pucicu: what parameters should be included?

* Mean recurrence time
* Recurrence time entropy
* Maximum recurrence time
* ... other?

On the other hand, what estimate to recurrence times do you recommend?
(a) distance between starting points of vertical recurrent structures
(b) length of vertical "white lines"

Both are easy to calculate and can be provided if necessary, but I wonder whether that would be complicating the interface too much...

pucicu:
The best would be to use the middle point of the vertical recurrence lines in the RP (see Fig. 1.4 in the book chapter N. Marwan, C. L. Webber, Jr.: Mathematical and Computational Foundations of Recurrence Quantifications, Recurrence Quantification Analysis – Theory and Best Practices, Eds.: C. L. Webber, Jr. and N. Marwan, Springer, Cham, 3–43 (2015). DOI:10.1007/978-3-319-07155-8_1 <http://dx.doi.org/10.1007/978-3-319-07155-8_1>). This corresponds more or less to an average of (a) and (b) – but it is not really the average, therefore, just averaging of both would not  work.

Regarding the other question: I suggest to include Mean recurrence time, Recurrence time entropy, but also the most probable recurrence time.

Thanks for your great work! It is really exciting to see this rapid progress!

pucicu:
hm, good question. I haven’t thought about this. perhaps just average the both values? would make sense, in my opinion.

heliosdrm:
I'm about finishing this. My main remaining doubt is about the names of the functions to calculate those parameters. Following [Julia's style guide](https://docs.julialang.org/en/v1/manual/style-guide/index.html) I propose this:

* `meanrecurrencetime` , with key `"MRT"` in the corresponding dictionary item returned by `rqa`.
* `recurrencetime_entropy`, with key `"RTE`".
* `mostprobable_recurrencetime`, with key `"NMPRT"`.

The "N" in the last key comes is present in the acronym given in Ngnamga's paper, standing for "number of the most probable recurrence time" (NMPRT), which is a more accurate description. (To be consistent the function name should be something like `"number_mostprobable_recurrencetime"`, but I think that it is just too much...

[...]

PR #31, opened 2019-01-06
[WIP] Rework test suite

Datseris:
The current tests have some problems. They test only a very specific trajectory of a very specific system. Although we both have no doubt that the package works as claimed, these tests can be seem as kind of fishy. 

They also were testing _exact_ values of RR, DET, etc. These kind of tests work only for the exact trajectory that was in the tests and not for the Lorenz system as a whole. So if one changed the integration window even slightly the tests would fail.

---

This PR tries to generalize the test suit so these test can be extended to many dynamical systems. For this we can use `DynamicalSystemsBase`, which offers many systems like e.g. `Systems.roessler`. By providing a vector of systems and a vector of test values one can test many systems with the same suite.

I have also replaced all `==` tests with `\approx`.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,0 +1,1 @@
+DynamicalSystemsBase 1.1
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,80 +1,54 @@
 using RecurrenceAnalysis
-using Test
+using Test, DynamicalSystemsBase
 
-# Functions for dynamic systems
-# Increment of x[ix] through RK4-integration
-# dxdh is the array of derivatives, dt is the time increment
-function dx_rk4(dxdt, x, ix, dt)
-    x = collect(x)
-    incr = zeros(length(x))
-    k1 = dxdt[ix](x...)
-    incr[ix] = dt*k1/2
-    k2 = dxdt[ix]((x+incr)...)
-    incr[ix] = dt*k2/2
-    k3 = dxdt[ix]((x+incr)...)
-    incr[ix] = dt*k3
-    k4 = dxdt[ix]((x+incr)...)
-    dt/6*(k1 + 2k2 + 2k3 + k4)
+systems = (Systems.lorenz(),)
+
+testvalues = (
+[426/234^2, 364/426, 364/50, 23, 2.02058292, 3.21794972e-7, 92/426, 92/30, 4],
+)
+
+u0s = (ones(3),)
+
+dt = 0.01
+for (ds, vals, u0) in zip(systems, testvalues, u0s)
+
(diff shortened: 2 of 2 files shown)


Datseris:
@heliosdrm At the moment the tests fail, even though I've tried to keep the trajectory as similar as possible with the original one. The current integration is significantly more accurate that then one used in the old tests.

Datseris:
Also, the lines:
```
    # Look for optimal threshold
    dd, rr = sorteddistances(x, theiler=1)
```
do not seem to be tested/used further

heliosdrm:
I'm also working on other (smaller) tests for #24 based on "hand-made" cross-recurrence matrices to verify that the RQA parameters are exact. I'll combine that with this, and use the Lorenz, Rossler, etc. time series only to check equalities and inequalities between different ways of making the calculations, so that the way they are integrated does not matter.

Datseris:
Awesome that is great! These test will then be the best possible, since they will test both hand made stuff _and_ real world applications!

src/matrices.jl
```diff
@@ -282,7 +282,7 @@ See also: [`CrossRecurrenceMatrix`](@ref).
 """
 function JointRecurrenceMatrix(x, y, ε; kwargs...)
     n = min(size(x,1), size(y,1))
-    rm1 = RecurrenceMatrix( (@view x[1:n,:]), ε, kwargs...)
-    rm2 = RecurrenceMatrix( (@view y[1:n,:]), ε, kwargs...)
+    rm1 = RecurrenceMatrix( (x[1:n,:]), ε, kwargs...)
```
Datseris: Here we should add an if clause:
```
if n == size(x, 1) && n == size(y, 1)
 RecurrenceMatrix(x, ...)
 ...
else
 # the code that exists now
end
```
to avoid making a copy in the most common case.

heliosdrm:
The new test sets are:

`dynamicalsystems.jl`: short time series: sine wave, random signal, and two examples of Hénon maps (one chaotic and another periodic) taken from a book. These tests verify that the functions do not throw errors, and some equivalences and inequalities between matrices and RQA parameters, including windowed RQA.

`smallmatrix.jl`: compares histograms of recurrence structures and RQA parameters calculated for a hand-made small matrix, against the expected exact values (manually calculated counting points and line one-by-one).

Includes a small fix on a bug of `JointRecurrenceMatrix`, detected with these new tests.

Datseris:
Yeah I have to admit these test are amazing! They exceed any possible standard I could set!

src/matrices.jl
heliosdrm: Not exactly: the old code works if `x` or `y` is `AbstractArray`, but not if it is a `Dataset`.
So, better make two different functions - depending on the type.

PR #32, opened 2019-01-06
lmin must be 1 or greater

heliosdrm:
the default is 2, but there is no reason to disallow `lmin==1`.


Datseris:
But doesn't `lmin = 1` means that single individual points on the recurrence plot are counted? Does this make sense?

heliosdrm:
Isolated recurrent points are recurrences anyway, so it does make sense. It's usual to count at least 2 points to consider it a recurrent "trajectory", that's why the RQA functions have a default `lmin=2`. But they work with any positive number of points. So, if for any reason a user conciously wants to decrease the limit, I think that they should be allowed to do it.

PR #42, opened 2019-01-16
zero RQA parameters with warning for empty histograms

### LINKED ISSUE #41: Parameters of "empty" histograms
This may happen more frequently in vertical structures, if there are no laminar states (no vertical lines, specially if a minimum line length is defined). But it might happen also with some signals for diagonal lines and recurrence times.

In such case the calculation of various parameters (averages, entropies, etc.) gives `NaN`. But is that the "right" solution? The condition is easy to check, so we may set them to zero in those cases if it's preferable. (Perhaps with a warning?)

heliosdrm:
Closes #41 

A problem is that the warning condition is hit several times during the tests (due to the examples of periodic signals without stationary states and in the windowed analysis).

Maybe it is not necessary to provide warnings. Actually there is nothing wrong with the null results; they are the correct result. Perhaps we could replace `@warn` by `@info`, to just highlight what is happening.

### DIFF
--- a/src/histograms.jl
+++ b/src/histograms.jl
@@ -138,7 +138,6 @@ function diagonalhistogram(x::ARM; lmin::Integer=2, theiler::Integer=0, kwargs..
             dh[1:nbins_loi] .+= loi_hist
         end
     end
-    (dh==[0]) && @warn "no diagonal lines found in the matrix"
     return dh
 end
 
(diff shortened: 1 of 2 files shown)


Datseris:
I think this is your call. How important is this to be mentioned and how should be in the hands of a person that has worked on the field (and I am definitely not qualified!). I am fine with all three ways, 0.0, `@warn` or `@info`. 

Maybe `0` is just the cleanest, since it doesn't do all these `(h == [0])` comparisons.

PR #50, opened 2019-01-18
Make `show` use UnicodePlots instead

Datseris:


Datseris:
```julia
using RecurrenceAnalysis
using DynamicalSystemsBase
using SparseArrays
using UnicodePlots

ro = Systems.roessler(a=0.15, b=0.20, c=10.0)
N = 2000; dt = 0.05
tr = trajectory(ro, N*dt; dt = dt, Ttr = 10.0)
R = RecurrenceMatrix(tr, 5.0; metric = "euclidean")

#prepare scatter data
function scatterdata(R)
   rows = rowvals(R)
   is = zeros(Int, nnz(R))
   js = zeros(Int, nnz(R))
   k = 1;
   m, n = size(R)
   for j = 1:n
     for i in nzrange(R, j)
        is[k] = j
        js[k] = rows[i]
        k += 1
     end
   end
   return is, js
end

is, js = scatterdata(R)
n, m = size(R)

a = UnicodePlots.scatterplot(
is, js, xlim = [1, n], ylim = [1, m], title = summary(R), labels = false,
color = :white, width = 80, height = 40)
```
shows in the Julia console in atom:
image

The operations are pretty much instant for a 2000x2000 and we can overload the width of the plot to be e.g. half of the total width available to the display used.

**This is mindblowing stuff**.

You can also have an ascii version that works everywhere forever and ever:

```julia
a = UnicodePlots.scatterplot(
is, js, xlim = [1, n], ylim = [1, m], title = summary(R), labels = false,
color = :white, width = 60, height = 30, border = :ascii, canvas = DotCanvas)
```
```
     RecurrenceMatrix of size (2001, 2001) with 513975 entries:
    +------------------------------------------------------------+
    |                 .:  .:'    .:' .:'        .:'.::'    .. .::|
    |               .:' .:'    .:' .:'        .:'.::'    .:'.::' |
    |             .:: .:'    .:: .:'    .:'..:'.::'   .::'.::'.: |
    |       .:: .::'.:' .:' ::'.:'  '  :'.::'.::'  '  :'.::'.:'  |
    |'.::'.::'.:::::'..:'    .:' .:    .::'.::'       .::'.::    |
    |''' ''' ''' ''' ''    .:'  ''..  .:'' ''  ' .. .::' '''   .:|
    |                ..  .:'    .:' .:'        .:'.::'    .  .::'|
    |                ' .:'..: ..' .:'..:     .:'.::'.:     .::'.:|
    |            ..  .:'.::'.:' .:' ::'   ..:'.::'.:' .  .::'.:' |
    |      ..  .:'..:'.::'   .::'.::'   .::'.::'.:'    .::'.:'   |
    |... .:::.:::::'.::'    .:'.::'    ::'.::'.:'     ::'.::     |
    |::.::::::::::.::'..     .::'       .::'.::'       .::'.:    |
    |:::::::':::'::' ::'   .::'       .::' ::'   ..  .::'.:'    .|
    |                    .::'    .:'.::'       .::'.:'        .:'|
    |                ''.::'..  .:'.::'.      .::'.:' .    ' .:' .|
    |             .: .::'.:' .:'.::'.:'    .::'.:' .:'.:' .:' .:'|
    |       .  .::'.::'.::  ''.::'.:'    .::'.:' .''  ' .:'..:'  |
    | .. ..:'.::'.::'.::'   .::'.:'    .::'.::'.:'    .::.::'    |
    | '   '  ''  ''  ''.  .::'  ' .: .::'  ''  '.:: .:'' ''    .:|
    |                .:'.::'   ..:'.::'       .::'.:'    .:  .:' |
    |             ..:'.::'.. .::'.::'   ..  .::'.:'    .:' .:' ..|
    |        .: .::'.::'.:' .:'.::' :  ::'.::'.:' .:  ::'.:' .:' |
    |  .::..:'.::'.::'.:'    .::'..     .::'.::       ..:'..:'   |
    |::::::::::'.::'.::'    ::'.:'     :::.::'.:      :::::'     |
    |::::::::::::'.::'       .::'      .::::::'       .::'       |
    |::::::::::::::'.:'     ::'        :::::'         ::.::      |
    |::::::::::::::::        .:'       .::::::        .::'       |
    |:::::::::::::::.        :         :::::'         :'.        |
    |:::::::::::::::'        .         ::::.          .::        |
    |::::::::::::::          '         :::::          :'.        |
    +------------------------------------------------------------+
```

heliosdrm:
Wow!

Datseris:
Do you think we should go with this being the default way to show the matrices? If yes I can replace the current `show`. If not, I think it is probably worth it to make this an available method. `UnicodePlots` is an extremely light dependency, which has only `StatsBase` in its dependencies (both are pure Julia and I Cannot imagine any user not having `StatsBase` in their packages anyway)

heliosdrm:
It's tempting. But I would like to sleep over it before deciding on whether `show` should do this. A couple of issues that we should think about:

* It's awesome and really good-looking, but might it be a burden in interactive use (regarding screen space taken, scrolling up-down after printing...)?

* Does this affect other interfaces outside the REPL, e.g. Juno, and how?

heliosdrm:
Ok, I love it. On with the `show`! :smile:

Datseris:
This is updated to have proper show for both short and verbose syntax. This is how it looks like:

image

@heliosdrm could you please checkout this PR as well and see how it looks on your end. There are two things I consider:

1. The name of the function `unicode_arm` (which currently sucks)
2. The default values for `minh` and `maxh`. I would like some feedback on those.

Datseris:
unicode vs. ascii: 

image

pucicu:
great! I like it very much!

PR #51, opened 2019-01-19
default theiler=1; 0 for cross-recurrence matrices

### LINKED ISSUE #45: [Doc] Theiler = 0
Does the value of `0` for the theiler window exclude the LOI or it includes it?

heliosdrm:
Closes #45 

It was a bit more difficult than what I expected, because redefining the methods for `CrossRecurrenceMatrix` with only a change in keyword arguments led me to recursive calls!

Finally I have achieved it using a "dummy" type `_RM<:AbstractRecurrenceMatrix`

https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/65b23c6a147c27e591f1a6ed1ad1ce9e9041f2b5/src/matrices.jl#L104-L108

But perhaps there is a smarter solution that I missed...

### DIFF
--- a/src/histograms.jl
+++ b/src/histograms.jl
@@ -103,7 +103,7 @@ function _linehistograms(rows::T, cols::T, lmin::Integer=1, theiler::Integer=0,
     return (bins, bins_d)
 end
 
-function diagonalhistogram(x::ARM; lmin::Integer=2, theiler::Integer=0, kwargs...)
+function diagonalhistogram(x::ARM; lmin::Integer=2, theiler::Integer=1, kwargs...)
     (theiler < 0) && throw(ErrorException(
         "Theiler window length must be greater than or equal to 0"))
     (lmin < 1) && throw(ErrorException("lmin must be 1 or greater"))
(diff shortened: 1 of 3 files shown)


Datseris:
Huh....

Isn't it just super simpler to define the function:
```julia
deftheiler(x) = 1
deftheiler(x::CrossRecurrenceMatrix) = 0
```
and in all functions replace _every_ `theiler = 0` with `thelier = deftheiler(x)`? Julia allows keywords to depend on previous arguments and the recurrence matrix is always an argument...

I believe the code became unnecessarily complicated, and this looks very ugly:
```
# Special type (not exported) to manage conditional parameters in methods
struct _RM <: AbstractRecurrenceMatrix
    data::SparseMatrixCSC{Bool,Int}
end
_RM(R::ARM) = _RM(R.data)
```

---

could you please why my simple suggestion isn't enough?

heliosdrm:
Yes, it is enough. I just was not smart enough to think on that! :flushed:

Datseris:
> Yes, it is enough. I just was not smart enough to think on that! 😳

Heh, no worries! Making this package took a lot of brainpower! These things are easy to overlook after writing so much code!

So we agree, right? Do you want me to do these changes?

heliosdrm:
Well, I did it already.

PR #56, opened 2019-01-21
clean code and docstrings + improve performance

heliosdrm:
`AbstractRecurrenceMatrix` and `textrecurrenceplot` are given in the documentation, so they should be exported.

Since the docstring of `rqa` goes first in the documentation pages, I have moved the full explanations of the behavior and conditional defaults of `theiler` there, and left only a reference to `rqa` in the rest of functions. (This also makes the text less repetitive.)

Explanations added about the ambiguities in the literature about the definition of recurrence times (in `recurrencestructures`) and  the denominator of the recurrence rate (in `recurrencerate`), with a rationale of the choices that we have made in this package.


heliosdrm:
The docstrings of the individual RQA functions are now expanded with more details about how they are computed (I hope I have not made mistakes with the latex formulas)

Datseris:
Awesome! I was going to open an issue for more docs in the RQA functions after we had 1.0 out, to not delay the release, but you did it already.

I will review this properly by building the latex documentation and checking all latex expressions when I am back home late night!

heliosdrm:
Made a change in `trend` which is not only documentation, but to gain performance.

That change makes the result slightly different for `CrossRecurrenceMatrix` made with time series of different lengths. But both results are consistent with the published formula, and I have not found results that I can compare in other software packages or fabricated examples. So perhaps we can set our own standard for such a rare case...

Or if we find such standard in the future and realize that the current computation is wrong for such cases, we fix the bug then.

PR #61, opened 2019-01-22
Change return value of rqa to named Tuple

### LINKED ISSUE #59: Return Type of rqa()
Why is the return type of rqa a Dict?
With a NamedTuple you still have the clear connection between the values and the name of the metric.
Furthermore, the return values are ordered, and it is a little bit cheaper because it needs less allocations.
See https://discourse.julialang.org/t/namedtuples-vs-dict/13119
I made some tests with NamedTuples, with the code on my branch. 
For a 100 x 100 RecurrenceMatrix the number of allocations is reduced from 257 to 209 and the running time is reduced from 112,5  microseconds to 107,8 microseconds. 
For me this is important, because I want to apply rqa on a large number of pixels (at least a million) with around 100 timesteps.

felixcremer:
This should be faster than using a Dict.
This would close #59

### DIFF
--- a/src/rqa.jl
+++ b/src/rqa.jl
@@ -346,27 +346,27 @@ Using this function is much more efficient than calling all individual functions
 one by one.
 
 ## Return
-The returned value is a dictionary with the following keys:
+The returned value is a NamedTuple with the following entries:
 
-* "RR": recurrence rate (see [`recurrencerate`](@ref))
-* "DET": determinsm (see [`determinism`](@ref))
-* "L": average length of diagonal structures (see [`dl_average`](@ref))
-* "Lmax": maximum length of diagonal structures (see [`dl_max`](@ref))
-* "DIV": divergence (see [`divergence`](@ref))
-* "ENTR": entropy of diagonal structures (see [`dl_entropy`](@ref))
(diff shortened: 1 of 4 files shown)


Datseris:
Could you please modify the documentation string? The return type is not a dictionary anymore and the keys are not strings. You can change e.g. `"RR"` to be surrounded by ticks instead of quotes (" -> `).

We also have to modify all test files since now we can't access `rqa_params` with `["RR"]` but with `.RR` instead (or with `[:RR]`)

felixcremer:
I will change the docstring and the tests tomorrow morning. 
Do we also need to change something in windowed.jl to make the return type consistent?
If so, I think I would need your assistance, because I don't understand the macro code.

Datseris:
Perfect, thanks!

> If so, I think I would need your assistance, because I don't understand the macro code.

Huhu, me neither :D 

Don't worry, I am sure @heliosdrm can take care of any change that needs to be done in the `@windowed` code.

heliosdrm:
Ok, I'll do it tomorrow

felixcremer:
@heliosdrm  Could you have a look at the changes I did in  windowed.jl. 
It is now returning a NamedTuple of vectors. 
I am now changing the tests.

PR #62, opened 2019-01-22
renames

### LINKED ISSUE #58: On the plot of a recurrence matrix
As I am preparing some material for the upcoming Julia tutorial I have noticed some difficulties when actually making the plots of recurrence matrices. 

In my tutorial I will be use the Lorenz system as a case study. To explain a recurrence matrix I will be comparing a parameter where the lorenz system is periodic and a parameter where it is chaotic. Here is an example that shows both a recurrence plot and a trajectory for the two cases:

```julia
lor = Systems.lorenz()
figure(figsize = (12,8))

for (i, ρ) in enumerate((160.0, 28.0))
    set_parameter!(lor, 2, ρ)
    t, dt = i == 1 ? (50.0, 0.01) : (50.0, 0.01)
    tr = trajectory(lor, t; dt = dt, Ttr = 2000.0)
    
    x = tr[:, 2]
    subplot2grid((3,2), (0, i-1))
    plot((0:dt:t)[1:2000], x[1:2000], "k")
    title("ρ = $ρ, " * (ρ != 160.0 ? "not periodic" : "periodic"))
    
    subplot2grid((3,2), (1, i-1), rowspan = 2)
    R = RecurrenceMatrix(tr, 5.0)
    imshow(recurrenceplot(R), cmap = "binary_r",  extent = (0, t, 0, t));
    xlabel("X"); ylabel("Z"); 
end
```
which gives

figure_7

What I noticed is that the bottom left plot is extremely misleading. For this parameter value `ρ=160` the Lorenz system is periodic. In fact examining the Poincare surface of section reveals that the system has "period 3" in the Poincare map. In addition, by looking at the timeseries plot one can see that the period of the trajectory is around 1 unit of time. This means that there should be a diagonal line spanning the recurrence plot, starting from time 1 (notice that the plots above are in correct time units not index units).

This is not what I see. I investigated further, by comparing the `recurrenceplot` function with the function `scatterdata`, which gives exactly the same plot data but in the form of a scatter plot. Here is the result:

```julia
lor = Systems.lorenz()
set_parameter!(lor, 2, 160.0)
tr = trajectory(lor, 50.0; dt = 0.01, Ttr = 2000.0)

R = RecurrenceMatrix(tr, 5)
L = size(R)[1]

figure(figsize = (10, 5))
subplot(121)
plot(RecurrenceAnalysis.scatterdata(R)..., linestyle = "None", marker = "o", ms = 0.1, color = "black")
gca().set_aspect("equal"); xlim(1, L); ylim(1, L)
subplot(122)
imshow(recurrenceplot(R), cmap = "binary_r",  extent = (0, L, 0, L));
```
which gives

figure_10

In the above the axis are in index units, divide by 100 to get real time. Anyway, there is a massive visual difference between the two plots. **My first concern with this issue is that it should be mentioned in the docs that `imshow`-like plots using `recurrenceplot` can be misleading**. We should also export `scatterdata` and probably find a better name like `scatterrecurrenceplot` (becomes too long).

---

I now have a second, conceptual concern. In the above plot I will now zoom in the bottom left corner of both plots, to come closer to time=1 (i.e. index=100) which is the period of the orbit:

figure_10_zoom

Now you can see that the `imshow` plot actually has the same information as the scatter plot. (I never doubted that, but a new user can be greatly misled).

Anyway, **my second conceptual concern is: why are the off diagonals not continuous? Why do they have white interruptions?** Since the orbit is periodic exact diagonal should exist right? And they should be uninterrupted. I am asking so I can give a proper explanation of the event during the tutorial.

---

EDIT: I Think I have resolved the second concern. Doing
```julia
figure()
plot3D(columns(tr)..., lw = 0.1)
```
shows that the orbit thickens when the Z variable reaches the maximum value. Must be some kind of weird quasi-periodicity.

Then the width may actually exceed the threshold of distance of `5` I had in the original computation. Using `ε = 10` instead resolves the problem.

Datseris:
bit more concistent names:
```
recurrencetext, recurrencescatter, recurrenceplot
```
also exporting `recurrencescatter`, because I want to put this to the docs as well.

Closes #58

### DIFF
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -1,11 +1,14 @@
+using UnicodePlots
+export recurrencetext, recurrencescatter, recurrenceplot
 #########################################
 # Text-based plotting
 #########################################
 """
-    scatterdata(R::ARM) -> xs, ys
-Transform the data of a recurrence matrix to scatter data `xs, ys`.
+    recurrencescatter(R) -> xs, ys
+Transform the data of a recurrence matrix to scatter data `xs, ys`
+(which are the indices of the recurrence points).
 """
-function scatterdata(R::ARM) # TODO: scan every ÷100 elements depending on size?
+function recurrencescatter(R::ARM) # TODO: scan every ÷100 elements depending on size?
    rows = rowvals(R)
    is = zeros(Int, nnz(R))
    js = zeros(Int, nnz(R))
@@ -21,13 +24,10 @@ function scatterdata(R::ARM) # TODO: scan every ÷100 elements depending on size
    return is, js
 end
 
-using UnicodePlots
-export textrecurrenceplot
-
-textrecurrenceplot(R::ARM; kwargs...) = textrecurrenceplot(stdout, R::ARM; kwargs...)
+recurrencetext(R::ARM; kwargs...) = recurrencetext(stdout, R::ARM; kwargs...)
 
 """
-    textrecurrenceplot([io,] R; minh = 25, maxh = 0.5, ascii, kwargs...) -> u
+    recurrencetext([io,] R; minh = 25, maxh = 0.5, ascii, kwargs...) -> u
 
 Obtain a text-scatterplot representation of a recurrence matrix `R` to be displayed in
 `io` (by default `stdout`). The matrix spans at minimum `minh` rows and at maximum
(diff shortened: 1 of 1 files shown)


heliosdrm:
What about renaming `recurrencescatter` as `recurrencecoordinates` ?

Perhaps  `recurrencetext` and `recurrenceplot` might have more meaningful names as well. `recurrencetext` is not that much intuitive, and both are "plots" somehow...

Datseris:
Honestly, I very much dislike the fact that we include `recurrence` in the names of all these functions. The `recurrence` part should be implied anyway from the input type. 

Can we do better I wonder? Of course we have to avoid names like only `scatterplot` etc. that may conflict

heliosdrm:
I was thinking nearly the same while I was coming home. What about:

* `coordinates` instead of `recurrencescatterplot`
* `image` instead of `recurrenceplot`
* And `recurrenceplot` is left available for the Unicode representation, which is actually the only real "plot".

heliosdrm:
Or maybe `grayscale`?

Datseris:
> That was a lot of work! 👍

Naaah, mostly Atom's FInd All -> Replace All magic :)

PR #63, opened 2019-01-24
specialize _computescale for identical inputs

heliosdrm:
The methods for `mean` and `maximum` give identical results, but here they are faster when used by `RecurrenceMatrix`because half of the values need not be calculated.

The general method of `_computescale` will give different results when used by `RecurrenceMatrix`, because `distances` will only have half of the values contained in the distance matrix, excluding the main diagonal. But this will make it consistent with the values returned when `mean` and `maximum` are used.

The condition is `x===y`, not `x==y`, because it is faster (the slowest case for `x==y` is precisely when both vectors have the same values, which will always happen for `RecurrenceMatrix`; with `x===y` the identity will be found immediately).

### DIFF
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -224,10 +224,8 @@ function _computescale(scale::Function, x, y, metric)
     if x===y
         distances = zeros(Int(length(x)*(length(x)-1)/2))
         c = 0
-        @inbounds for (i,xi) in enumerate(x), (j,yj) in enumerate(y)
-            if j > i
-                distances[c+=1] = evaluate(metric, xi, yj)
-            end
+        @inbounds for i in 1:length(x)-1, j=(i+1):length(x)
+            distances[c+=1] = evaluate(metric, x[i], y[j])
         end
     else
         distances = distancematrix(x, y, metric)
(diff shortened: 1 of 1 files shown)


src/matrices.jl
```diff
@@ -219,23 +219,54 @@ end
...
+    return scale(distances)
+end
 # specific methods to avoid `distancematrix`
 function _computescale(scale::typeof(maximum), x, y, metric::Metric)
     maxvalue = zero(eltype(x))
-    @inbounds for xi in x, yj in y
-        newvalue = evaluate(metric, xi, yj)
-        (newvalue > maxvalue) && (maxvalue = newvalue)
+    if x===y
+        @inbounds for (i,xi) in enumerate(x), (j,yj) in enumerate(y)
```
Datseris: I wonder. Can't this loop be reduced even further, to not require 2 full loops? I think we can do something like:
```julia
for i in 1:length(x)-1  
  for j in (i+1):length(x)
    @inbounds newvalue = evaluate(metric, x[i], y[j])
     (newvalue > maxvalue) && (maxvalue = newvalue)
  end
end
```

I think the above avoids the boolean comparison alltogether. But Is it right? I just pulled out of the top of my head :P

heliosdrm:
yes, you are right! memory allocation is exactly the same, but improves a few ms in matrices ~ 5000x5000

Datseris:
Before I merge, I have question:

> The methods for mean and maximum give identical results

I don't get this part... Why would `mean` and `maximum` lead to the same result? The doc says:

> Typical choices are maximum or mean, such that the threshold ε is defined as a ratio of the maximum or the mean distance between data points, respectively (using mean or maximum calls specialized versions that are faster than the naive approach)

heliosdrm:
Sorry, I didn't explain it clearly... I meant "identical results" to the ones before the PR, i.e. that the change of code in those functions only changes the performance, not the result. `_computescale(mean, ...)` is as different from `_computescale(maximum, ...)` as it was before!

PR #72, opened 2019-02-10
test on Julia 1.0 (LTS)

heliosdrm:
I think Julia 1.0 should be restored in the tests, because that's the long term support version of Julia, and packages should aim at providing LTS as well. If that test breaks in the future, a branch to support 1.0 can be created and maintained separately.


Datseris:
I agree.

Issue #77, opened 2019-09-19
Clean up source a bit


Datseris:
It is possible to clean up and re-organize the source code a bit more for readability and consiseness . A lot of development was put into releasing 1.0 but since then I didn't really put effort to "clean things up".

Datseris:
(especially in the file `matrices.jl`)

Datseris:
While I am at it, also simplify expressions, e.g.
`nzvals = fill(true, (length(rowvals),))` to `nzvals = trues(length(rowvals))`

PR #81, opened 2019-09-25
Parallelize recurrence matrix construction

### LINKED ISSUE #76: Performance of computation of recurrence matrices

```julia
    x = xx.data
    y = yy.data
    rowvals = Vector{Int}()
    colvals = Vector{Int}()
    for j in 1:length(y)
        nzcol = 0
        for i in 1:length(x)
                push!(rowvals, i)
                nzcol += 1
            end
        end
    end
end
```

asinghvi17:
Adds a kwarg to `_recurrence_matrix`, which controls whether the 
parallel or serial algorithm is called.

Partially fixes #76

### DIFF
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -2,6 +2,7 @@ module RecurrenceAnalysis
 
 using Distances, Statistics, LinearAlgebra, SparseArrays, DelayEmbeddings, StaticArrays
 import Base.Meta.parse
+using Base.Threads
 
 export RecurrenceMatrix, CrossRecurrenceMatrix, JointRecurrenceMatrix,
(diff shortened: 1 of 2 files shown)


Datseris:
@asinghvi17 I am sorry, but could you please wait to progress with this after I do #77 ? I will be simplifing the `recurrencematrix` interface there. And then this may also need less work to be done.

asinghvi17:
It's pretty much done,  just needs the parallelization heuristic.

Datseris:
Hi @asinghvi17 , can you please adapt this based on #82 ?

asinghvi17:
Will look into it today.

Datseris:
Oh Cool! SORRY i forgot about this!

Issue #22, opened 2018-12-31
Specific code for `recurrencematrix`


heliosdrm:
`recurrencematrix` calls `crossrecurrencematrix` with the same `x` and `y` arguments, thus duplicating the number of comparisons and the size of the nonzero indices of the resulting matrix.

If we have different `RecurrenceMatrix` and `CrossRecurrenceMatrix` types, the size and number of calculations for the former may be reduced. Type-specific methods for RQA and for the creation of the plottable matrix can take into account the symmetry of recurrence matrices. (And also for joint recurrence matrices.)

Datseris:
Yeah, I guess it is better to do this via multiple dispatch than by comparing strings :D I'll edit #25

heliosdrm:
So, are you working on this at the same time as in the new types?
This is not breaking, but only to improve performance, so perhaps you may finish #25 first without dealing with symmetry, and then we put our hands into this.

This needs:
- [ ] Code a specific `recurrencematrix` method that only populates half of the matrix
- [ ] Change `recurrencerate`, `diagonalhistogram` and `verticalhistogram`, to dispatch on specific types.

For the second task, there may be one method that dispatches on `Union{`RecurrenceMatrix`, `JointRecurrenceMatrix`} (symmetric), and another for `CrossRecurrenceMatrix`. The current code uses one method for all, branching according to the result of `issymmetric`.

I can take over this when #25 is finished.

Datseris:
> This is not breaking, but only to improve performance, so perhaps you may finish #25 first without dealing with symmetry, and then we put our hands into this. 

Yeap, this is the plan for now.

> I can take over this when #25 is finished.

Awesome, thanks so much!

PR #82, opened 2019-10-01
Simplify code for `recurrence_matrix`

### LINKED ISSUE #60: Specialized method for `_recurrence_matrix(::Vector, ::Vector)`
Currently all recurrence_matrix calls transform data to Datasets in all cases. This is always faster for Matrices but I do not believe that this does not induce a performance regression for vectors.

A method should be added to the low level function `_recurrence_matrix` that does not do this transformation and also does not care about the metric and instead uses `abs(xi - yi)`

(notice: to get the gist of this issue is straightforward, but requires you to have a look at the source file:  https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/master/src/matrices.jl

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/69047699-specialized-method-for-_recurrence_matrix-vector-vector?utm_campaign=plugin&utm_content=tracker%2F28405688&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F28405688&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #77: Clean up source a bit

Datseris:
This is my attempt at simplification. Unfortunately, I didn't go far. I was only able to do very small changes.

The main difficult is the processing we do for fixed /nonfixed scale. Usually you want to have only two levels: a high level API and a low level implementation. You change the high level for documentation and clarity, and you change the low level for performance. The difficulty here is that we have the intermediate level, the code that does the processing for the scaling. 

My dream goal would be to remove the function `recurrence_matrix` alltogether, so that only `_recurrence_matrix` exists, but this seems to be too difficult to achieve.

Regardless, using `Base.front()` I was able to at least modify it so that `_recurrence_matrix(x, ε)` and `_recurrence_matrix(x, y, ε)` are still directly callable from the high-level API. This paves the way for #22 .

Closes #77 .

(while I was doing that I also added #60 )

Closes #60 .

### DIFF
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -3,8 +3,28 @@ module RecurrenceAnalysis
 using Distances, Statistics, LinearAlgebra, SparseArrays, DelayEmbeddings, StaticArrays
 import Base.Meta.parse
 
+const METRICS = Dict(
+    "euclidean"=>Euclidean(),
+    "max"=>Chebyshev(),
+    "inf"=>Chebyshev(),
+    "cityblock"=>Cityblock(),
+    "manhattan"=>Cityblock(),
+    "taxicab"=>Cityblock(),
+    "min"=>Cityblock()
+)
+const DEFAULT_METRIC = Euclidean()
+getmetric(m::Metric) = m
+function getmetric(normtype::AbstractString)
+    normtype = lowercase(normtype)
+    !haskey(METRICS,normtype) && error("incorrect norm type. Accepted values are \""
+        *join(keys(METRICS),"\", \"", "\" or \"") * "\".")
+    METRICS[normtype]
+end
+
+
 export RecurrenceMatrix, CrossRecurrenceMatrix, JointRecurrenceMatrix,
        AbstractRecurrenceMatrix
+
 export embed,
        reconstruct,
        Dataset,
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -1,25 +1,14 @@
-const METRICS = Dict(
-    "euclidean"=>Euclidean(),
-    "max"=>Chebyshev(),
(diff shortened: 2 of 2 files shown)


Datseris:
I see there already optimizations if `x === y` in the source:
```
if x===y
        distances = zeros(Int(length(x)*(length(x)-1)/2))
        c = 0
        @inbounds for i in 1:length(x)-1, j=(i+1):length(x)
            distances[c+=1] = evaluate(metric, x[i], y[j])
        end
    else
        distances = distancematrix(x, y, metric)
    end
```

Datseris:
There more I think about it, I think I am just overzealous for this. Both the scaling thing and the `_recurrence_matrix` are low level and completely independent from each other, so I think it is fine.

What I will do in a moment is to ensure they are not chained: both should be called form the highest level, e.g. `RecurrenceMatrix`, etc.

Datseris:
Ah important: this changes the order between `ε, metric` in the lowest level (non-breaking at high level of course).

Issue #83, opened 2019-12-30
Warning: implicit `dims=2` argument now has to be passed explicitly to specify that distances between columns should be computed


yuxiliu1995:
During testing with `pkg > test RecurrenceAnalysis`, the test threw this warning:
```
┌ Warning: implicit `dims=2` argument now has to be passed explicitly to specify that distances between columns should be computed
│   caller = ip:0x0
└ @ Core :-1
```

Datseris:
Okay, but what is the line that throws this?

yuxiliu1995:
I don't know. The full printout is 

```
(v1.3) pkg> test RecurrenceAnalysis
   Testing RecurrenceAnalysis
 Resolving package versions...
    Status `C:\Users\DEADSC~1\AppData\Local\Temp\jl_urrXqk\Manifest.toml`
  [4fba245c] ArrayInterface v2.3.1
  [9e28174c] BinDeps v1.0.0
  [b99e7846] BinaryProvider v0.5.8
  [bbf7d656] CommonSubexpressions v0.2.0
  [34da2185] Compat v2.2.0
  [9a962f9c] DataAPI v1.1.0
  [864edb3b] DataStructures v0.17.6
  [5732040d] DelayEmbeddings v1.2.0
  [2b5f629d] DiffEqBase v6.10.0
  [01453d9d] DiffEqDiffTools v1.7.0
  [163ba53b] DiffResults v1.0.2
  [b552c78f] DiffRules v1.0.0
  [b4f34e82] Distances v0.8.2
  [ffbed154] DocStringExtensions v0.8.1
  [6e36e845] DynamicalSystemsBase v1.4.0 [`C:\Users\DeadScholar\.julia\dev\DynamicalSystemsBase`]
  [f6369f11] ForwardDiff v0.10.8
  [069b7b12] FunctionWrappers v1.0.0
  [42fd0dbc] IterativeSolvers v0.8.1
  [82899510] IteratorInterfaceExtensions v1.0.0
  [1914dd2f] MacroTools v0.5.3
  [e1d29d7a] Missings v0.4.3
  [46d2c3a1] MuladdMacro v0.2.1
  [77ba4419] NaNMath v0.3.3
  [b8a86587] NearestNeighbors v0.4.4
  [bac558e1] OrderedCollections v1.1.0
  [d96e819e] Parameters v0.12.0
  [3cdcf5f2] RecipesBase v0.7.0
[...]

PR #84, opened 2019-12-31
Update compat bounds of dependencies and tests

heliosdrm:
Fix depwarn of Distances v0.8, cf. #83


heliosdrm:
Do not merge yet, there is something failing...

Datseris:
This should also add a compat entry of `Distances = 0.8` then, right?

heliosdrm:
You're right. It's also updated.

By the way, maybe we should also modify the `Project.toml` in order to specify upper bounds to all compats (including Julia), following the recommended policy of the General Registry...

I'l file an issue about that.

Datseris:
> By the way, maybe we should 

Yeap I have done this in the other repo's, we should do it here as well. In fact, this PR is a good idea to do it in.

heliosdrm:
Ok, I have changed the title of the PR and will add here the commits for Project.toml

heliosdrm:
Also added tests for OSX and Windows in Travis, but if you think that this makes testing too heavy, it can be removed.

heliosdrm:
Julia 0.7 is not happy with the current code. Shall we drop 0.7 support?

Datseris:
> Julia 0.7 is not happy with the current code. Shall we drop 0.7 support?

Yeap go ahead! You can also remove AppVeyor, we now test windows on travis anyway!

Issue #67, opened 2019-01-26
[FR] Estimations of `ε` 


Datseris:
Would be great if we had a function with various methods to estimate `ε` , the threshold of a recurrence plot.

This could work similarly with `estimate_delay` which gives delay times for timeseries.

Different methods are described in the book from N. Marwan and citations within, see chapter 1 (we cite the book in most docs). Some of these methods are actually very easy and straightforward and would be a great issue for newcomer contributor.

Datseris:
Awesome, please do so!

Datseris:
This issue now has a 100$ bounty on it. (Which means anyone who solves it gets the 100$)

(edit: not anymore)

Datseris:
related: https://aip.scitation.org/doi/10.1063/1.5024914

yuxiliu1995:
Note: The https://aip.scitation.org/doi/10.1063/1.5024914 paper is extremely wordy, and basically says that "counting the shortest N% of distances in the recurrence plot as recurring" is a good way to ensure some Recurrence Quantification Analysis less sensitive to embedding dimensions.

yuxiliu1995:
I read chapter 1 of Webber and Marwan. It seems that this question is solved? Just pass kwarg to the integrator, and it's done. The only notable 

Thus possibly this bounty is outdated.

Datseris:
@yuxiliu1995 can you please give more detail? What we are talking about here is having a method (a function) that you give in your dataset and then function returns an optimal value for `ε`, much like the functions `estimate_delay` which return an optimal value for e.g. `τ`.

yuxiliu1995:
I read Chapter 1 of Marwan, https://aip.scitation.org/doi/10.1063/1.5024914 , as well as many cited papers. It seems there is no good way to do this other than to try a few and see what works. It is just a collection of gut feelings, case studies, and appeal to numerical simulation.

Not only is it unmathematical, it is also simple. There are only a few methods for choosing ε:

* `ε = p * std`, with `0.2 < p < 0.4`
* `ε < 0.1 * mean`.
* `ε = p * median` with `p ~ 0.2`.
* `ε = p * diameter`, with `p < 0.1`, or `p ~ 0.05`, or `0.2 < p < 0.4`.
* `ε = p-quantile`, with `0.01< p < 0.05`.

* `ε > 5 * std of noise`, in the case where the measurement device has a Gaussian noise distribution with a known std.
[...]

PR #89, opened 2020-02-12
Estimations of `ε`

yuxiliu1995:
https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/issues/67

### DIFF
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -1,3 +1,5 @@
+using Statistics
+using ChaosTools:linear_region
 #=
 In this file the core computations for creating a recurrence matrix
 are defined (via multiple dispatch).
@@ -169,7 +171,18 @@ Quantification Analysis. Theory and Best Practices*, Springer, pp. 3-43 (2015).
 """
 function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC, kwargs...)
     m = getmetric(metric)
+<<<<<<< HEAD
     s = resolve_scale(x, m, ε; kwargs...)
+=======
+    if !haskey(kwargs, :scale)
+        s = resolve_scale(x, x, m, ε; kwargs...)
+    elseif kwargs[:scale] == "powerlaw"
+        s = _powerlawepsilon(x, x; kwargs...)
+    elseif kwargs[:scale] == "peak"
+        s =_significantpeaksepsilon(x, x; kwargs...)
+    end
+
+>>>>>>> implement several methods for epsilon selection
     m = recurrence_matrix(x, m, s)
     return RecurrenceMatrix(m)
 end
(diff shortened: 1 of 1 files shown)


src/matrices.jl
```diff
@@ -169,7 +171,18 @@ Quantification Analysis. Theory and Best Practices*, Springer, pp. 3-43 (2015).
 """
 function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC, kwargs...)
     m = getmetric(metric)
+<<<<<<< HEAD
```
Datseris: this has to be a mistake, along with the other `=====` stuff from resolving conflicts.

yuxiliu1995: fixed

yuxiliu1995:
Try this example usage:

```julia
using DynamicalSystems, PyPlot, RecurrenceAnalysis

R = RecurrenceMatrix(tr, 0.1; scale="powerlaw")
Rg = grayscale(R)
imshow(Rg, cmap = "binary_r", extent = (1, size(R)[1], 1, size(R)[2]))
```

Change `scale="powerlaw"` to `scale=median`, `scale=var`, `scale="peak"` for the other methods I implemented.

`"peak"` gives very poor results for some reason.

src/matrices.jl
```diff
@@ -1,3 +1,5 @@
+using Statistics
+using ChaosTools:linear_region
```
Datseris: You can't add `ChaosTools` as a dependency of RecurrenceAnalysis. You can just copy the `linear_region` code, but I think it is probably better to move `linear_region` to DelayEmbeddings.jl, as it is a more basic package and the method is generally useful. 

I'll do this ASAP.

Datseris:
Norbert added some comments in #67 that should be incorporated here.

src/matrices.jl
```diff
@@ -169,7 +171,14 @@ Quantification Analysis. Theory and Best Practices*, Springer, pp. 3-43 (2015).
 """
 function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC, kwargs...)
     m = getmetric(metric)
-    s = resolve_scale(x, m, ε; kwargs...)
+    if !haskey(kwargs, :scale)
+        s = resolve_scale(x, x, m, ε; kwargs...)
+    elseif kwargs[:scale] == "powerlaw"
```
asinghvi17: Should this be replaced with a type-based dispatch system?  I find that it's immensely helpful to specify options, and clarify dispatches...

Datseris: I am not sure how helpful this will be in the end of the road, because it would make the code much larger. How do you think it would be done to keep the code the same size?

In addition, from the user perspective `method = "linear"` or `method = "cubic"` is simpler to understand than `method = Linear()`, `method = Cubic()`. Everyone knows strings, but not everyone knows singleton types and their role in multiple dispatch.

But in general I also prefer using multiple dispatch instead of unending `if` statements.

yuxiliu1995:
btw, I have lost interest in this issue. It feels too experimental and theoretically shallow.
Anyone is free to take my code and continue.

PR #90, opened 2020-02-26
Parallelize recurrence plots v2.0

### LINKED ISSUE #76: Performance of computation of recurrence matrices

asinghvi17:
Parallelizes the recurrence plots by using Julia's multithreading.  

Supersedes #81; partially fixes #76

### DIFF
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -369,6 +369,7 @@ function recurrence_matrix(xx::Dataset, yy::Dataset, metric::Metric, ε, paralle
 
     # This is the same logic as the serial function, but parallelized.
     Threads.@threads for j in 1:length(y)
+        threadn = Threads.threadid()
         nzcol = 0
         for i in 1:length(x)
             @inbounds if evaluate(metric, x[i], y[j]) ≤ ε
@@ -394,6 +395,7 @@ function recurrence_matrix(x::AbstractVector, y::AbstractVector, metric, ε, par
 
     # This is the same logic as the serial function, but parallelized.
     Threads.@threads for j in 1:length(y)
+        threadn = Threads.threadid()
         nzcol = 0
         for i in 1:length(x)
             @inbounds if abs(x[i] - y[j]) ≤ ε
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -213,14 +213,14 @@ length, the recurrences are only calculated until the length of the shortest one
 See [`RecurrenceMatrix`](@ref) for details, references and keywords.
 See also: [`CrossRecurrenceMatrix`](@ref).
 """
-function JointRecurrenceMatrix(x, y, ε; parallel::Bool = false, kwargs...)
+function JointRecurrenceMatrix(x, y, ε; kwargs...)
     n = min(size(x,1), size(y,1))
     if n == size(x,1) && n == size(y,1)
-        rm1 = RecurrenceMatrix(x, ε; parallel = parallel, kwargs...)
-        rm2 = RecurrenceMatrix(y, ε; parallel = parallel, kwargs...)
+        rm1 = RecurrenceMatrix(x, ε; kwargs...)
+        rm2 = RecurrenceMatrix(y, ε; kwargs...)
     else
-        rm1 = RecurrenceMatrix(x[1:n,:], ε; parallel = parallel, kwargs...)
-        rm2 = RecurrenceMatrix(y[1:n,:], ε; parallel = parallel, kwargs...)
+        rm1 = RecurrenceMatrix(x[1:n,:], ε; kwargs...)
+        rm2 = RecurrenceMatrix(y[1:n,:], ε; kwargs...)
     end
     return JointRecurrenceMatrix(rm1.data .* rm2.data)
 end
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -171,7 +171,8 @@ use [`recurrenceplot`](@ref) to turn the result of these functions into a plotta
 recurrence quantifications", in: Webber, C.L. & N. Marwan (eds.), *Recurrence
 Quantification Analysis. Theory and Best Practices*, Springer, pp. 3-43 (2015).
 """
-function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC, parallel::Bool = false, kwargs...)
+function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC,
+    parallel::Bool = length(x) > 500 && Threads.nthreads() > 1, kwargs...)
     m = getmetric(metric)
     s = resolve_scale(x, m, ε; kwargs...)
     m = recurrence_matrix(x, m, s, Val(parallel))
@@ -191,7 +192,8 @@ then the cell `(i, j)` of the matrix will have a `true` value.
 See [`RecurrenceMatrix`](@ref) for details, references and keywords.
 See also: [`JointRecurrenceMatrix`](@ref).
 """
-function CrossRecurrenceMatrix(x, y, ε; metric = DEFAULT_METRIC, parallel::Bool = false, kwargs...)
+function CrossRecurrenceMatrix(x, y, ε; metric = DEFAULT_METRIC,
+    parallel::Bool = length(x) > 500 && Threads.nthreads() > 1, kwargs...)
     m = getmetric(metric)
     s = resolve_scale(x, y, m, ε; kwargs...)
     m = recurrence_matrix(x, y, m, s, Val(parallel))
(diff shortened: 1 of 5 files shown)


asinghvi17:
```
julia> @benchmark RecurrenceMatrix($(tr), $(5.0); metric = "euclidean", parallel = false)
BenchmarkTools.Trial:
  memory estimate:  20.29 MiB
  allocs estimate:  2056
  --------------
  minimum time:     21.966 ms (0.00% GC)
  median time:      27.255 ms (0.00% GC)
  mean time:        26.889 ms (7.89% GC)
  maximum time:     37.462 ms (15.30% GC)
  --------------
  samples:          186
  evals/sample:     1

julia> @benchmark RecurrenceMatrix($(tr), $(5.0); metric = "euclidean", parallel = true)
BenchmarkTools.Trial:
  memory estimate:  32.76 MiB
  allocs estimate:  2171
  --------------
  minimum time:     14.249 ms (0.00% GC)
  median time:      15.865 ms (0.00% GC)
  mean time:        18.339 ms (14.33% GC)
  maximum time:     30.549 ms (47.84% GC)
  --------------
  samples:          273
  evals/sample:     1
```

Datseris:
What's the size of `tr` here?

Datseris:
Do you think you could do a simple graphf of speed up gain versus dataset size? maybe we could add a simple heuristic like:

if dataset size < N && nthreads() > 1  then do parallel

src/matrices.jl
```diff
@@ -155,22 +156,25 @@ by the following keyword arguments:
...
+[^2] : N. Marwan & C.L. Webber, "Mathematical and computational foundations of
 recurrence quantifications", in: Webber, C.L. & N. Marwan (eds.), *Recurrence
 Quantification Analysis. Theory and Best Practices*, Springer, pp. 3-43 (2015).
 """
-function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC, kwargs...)
+function RecurrenceMatrix(x, ε; metric = DEFAULT_METRIC, parallel::Bool = false, kwargs...)
     m = getmetric(metric)
     s = resolve_scale(x, m, ε; kwargs...)
-    m = recurrence_matrix(x, m, s)
+    m = recurrence_matrix(x, m, s, Val(parallel))
```
Datseris: Does using `Val` result into performance gain, given the cool optimizations of the latest Julia version? I know it makes it "more type stable", but is it actually worth it? (because it sacrifices code clarity)

src/matrices.jl
```diff
@@ -209,14 +213,14 @@ length, the recurrences are only calculated until the length of the shortest one
 See [`RecurrenceMatrix`](@ref) for details, references and keywords.
 See also: [`CrossRecurrenceMatrix`](@ref).
 """
-function JointRecurrenceMatrix(x, y, ε; kwargs...)
+function JointRecurrenceMatrix(x, y, ε; parallel::Bool = false, kwargs...)
     n = min(size(x,1), size(y,1))
     if n == size(x,1) && n == size(y,1)
-        rm1 = RecurrenceMatrix(x, ε; kwargs...)
-        rm2 = RecurrenceMatrix(y, ε; kwargs...)
+        rm1 = RecurrenceMatrix(x, ε; parallel = parallel, kwargs...)
```
Datseris: Since `parallel` is used everywhere anyway, should you just propagate it along with the rest of `kwargs...` ?

asinghvi17:
```julia
using DynamicalSystems
using DynamicalSystems
ro = Systems.roessler(ones(3), a=0.15, b=0.20, c=10.0)
N = 2000; dt = 0.05
tr = trajectory(ro, N*dt; dt = dt, Ttr = 10.0)

@benchmark RecurrenceMatrix($(tr), $(5.0); metric = "euclidean", parallel = false)
BenchmarkTools.Trial:
  memory estimate:  20.29 MiB
  allocs estimate:  2056
  --------------
  minimum time:     21.966 ms (0.00% GC)
  median time:      27.255 ms (0.00% GC)
  mean time:        26.889 ms (7.89% GC)
  maximum time:     37.462 ms (15.30% GC)
  --------------
  samples:          186
  evals/sample:     1

#

@benchmark RecurrenceMatrix($(tr), $(5.0); metric = "euclidean", parallel = true)
BenchmarkTools.Trial:
  memory estimate:  32.76 MiB
  allocs estimate:  2171
  --------------
  minimum time:     14.249 ms (0.00% GC)
  median time:      15.865 ms (0.00% GC)
  mean time:        18.339 ms (14.33% GC)
  maximum time:     30.549 ms (47.84% GC)
  --------------
  samples:          273
  evals/sample:     1
```

and the coup de grace:

```julia
N = 10000; dt = 0.05
tr = trajectory(ro, N*dt; dt = dt, Ttr = 10.0)
@benchmark RecurrenceMatrix($(tr), $(5.0); metric = "euclidean", parallel = false)
BenchmarkTools.Trial:
  memory estimate:  482.13 MiB
  allocs estimate:  10067
  --------------
  minimum time:     731.655 ms (7.31% GC)
  median time:      803.185 ms (16.76% GC)
  mean time:        800.596 ms (13.89% GC)
  maximum time:     892.434 ms (15.20% GC)
  --------------
  samples:          7
  evals/sample:     1

@benchmark RecurrenceMatrix($(tr), $(5.0); metric = "euclidean", parallel = true)
BenchmarkTools.Trial:
  memory estimate:  597.54 MiB
  allocs estimate:  10200
  --------------
  minimum time:     473.195 ms (0.00% GC)
  median time:      572.614 ms (15.97% GC)
  mean time:        591.944 ms (18.67% GC)
  maximum time:     776.823 ms (34.54% GC)
  --------------
  samples:          9
  evals/sample:     1
```

src/matrices.jl
asinghvi17: Good point - will do!

Datseris:
Hm, I am surprised the speedup is so small for such a large dataset, 570 ms vs 800 ms. How many threads do you have? I'll also test this on my computer. Let's write a script that does this and plots speedup versus N or so.

asinghvi17:
I have 4 threads on my computer, but also running atom, powerpoint, etc.  Do let me know what it looks like on your machine!

asinghvi17:
A script would look something like:

```julia
using DynamicalSystems, BenchmarkTools, RecurrenceAnalysis

function ExpRange(lower, upper, size; base = ℯ)
    return base .^ LinRange(log(base, lower), log(base, upper), size)
end

function create_dataset(size; system = Systems.roessler(ones(3), a=0.15, b=0.20, c=10.0), dt = 0.05, Ttr = 10.0)
    return trajectory(system, size*dt; dt = dt, Ttr = Ttr)
end

sizes = ExpRange(10, 100_000, 10)
datasets = create_dataset.(sizes)

times = []

for dataset in datasets
    b = @benchmark RecurrenceMatrix($(tr), $(5.0); metric = Euclidean(), parallel = true)
    push!(times, median(b.times))
end
```

or something

asinghvi17:
For `n = 50_000`, I see 31s serial v/s 19s parallel, so a 1/3 speedup.

src/matrices.jl
asinghvi17: I actually think it increases code clarity, which is why I used it - we can separate implementation methods.  Later, if we choose to define a distributed or GPU backend for calculating recurrence matrices, it will become easier to implement if we're already dispatching on types as opposed to having an if statement with keyword arguments.

asinghvi17:
Blue is serial, orange is parallel:
lin-scene



log-scene

Datseris:
I got:
```
I am using 12 threads. I am reporting results as:
time of (serial/parallel) for 3D and 1D trajectories.
For N = 101...
3D=0.884, 1D=0.822
For N = 1001...
3D=1.49, 1D=1.25
For N = 10001...
3D=1.46, 1D=1.3
For N = 100001...
```

see the latest file I've uploaded. Seems pretty good :) I guess the heuristic could be at `N=500`?

Datseris:
I'll wait a couple of days if @heliosdrm want to comment otherwise we _mergerino_ this.!

src/matrices.jl
Datseris: Hm, okay sounds convincing.

asinghvi17:
1.3.1 on MacOS randomly times out, nothing we can do I think

heliosdrm:
Very nice job! Thank you @asinghvi17 !

Before merging, it would be nice to add a test script, e.g. an example like the one you have shown in a `test/parallelized.jl` (with data series that are not too large - it is for testing that it works, not benchmarking performance), and a line in `test/runtests.jl` including it.

`Project.toml` should be then updated to reflect a minor change (version 1.2.0), and maybe include new entries in the `[extras]` and `[targets]` sections, if new modules are loaded in the test.

Datseris:
I've added a simple heuristic for parallelization, let me know if it is fine. I'll merge once tests pass!

PR #91, opened 2020-03-21
Only calculate the lower triangle

### LINKED ISSUE #76: Performance of computation of recurrence matrices

asinghvi17:
This PR successfully implements a serial and parallel version of the single-trajectory recurrence matrix calculation.  By only calculating the lower triangle's recurrence, we can save on a significant amount of time.

In parallel, what I've done is to equalize the number of calculations per thread, by partitioning the array such that each thread has an equal "area" to iterate over.  If the whole iteration space is a triangle, then what I did here is to partition it into trapezoids of equal area.

Fixes #76!

### DIFF
--- a/benchmarks/parallel_vs_serial.jl
+++ b/benchmarks/parallel_vs_serial.jl
@@ -4,21 +4,71 @@ This file benchmarks the parallel and serial computation of a recurrence matrix
 using DynamicalSystemsBase, RecurrenceAnalysis
 using Statistics, BenchmarkTools, Base.Threads
 
+function pretty_time(b::BenchmarkTools.Trial)
+    med = median(b)
+
+    return string(
+        BenchmarkTools.prettytime(BenchmarkTools.time(med)),
+        " (",
+        BenchmarkTools.prettypercent(BenchmarkTools.gcratio(med)),
+        " GC)"
+    )
+
+end
+
 Ns = round.(Int, 10.0 .^ (2:0.5:4.5))
 ro = Systems.roessler()
 
-println("I am using $(nthreads()) threads. I am reporting results as:")
-println("time of (parallel/serial) for 3D and 1D trajectories.")
+println("I am using $(nthreads()) threads.")
 for N in Ns
+    # set up datasets
     tr = trajectory(ro, N*0.1; dt = 0.1, Tr = 10.0)
-    println("For N = $(length(tr))...")
+    printstyled("For N = $(length(tr))..."; color = :blue, bold = true)
+    println()
     x = tr[:, 1]
+
+    # calculate only the lower triangle
     b_serial   = @benchmark RecurrenceMatrix($(tr), 5.0; metric = Euclidean(), parallel = false)
     b_parallel = @benchmark RecurrenceMatrix($(tr), 5.0; metric = Euclidean(), parallel = true)
     b_s_1 = @benchmark RecurrenceMatrix($(x), 2.0; parallel = false)
     b_p_1 = @benchmark RecurrenceMatrix($(x), 2.0; parallel = true)
+
+    # calculate triangular metrics
     threeD = median(b_serial.times)/median(b_parallel.times)
     oneD = median(b_s_1.times)/median(b_p_1.times)
     threeD, oneD = round.((threeD, oneD); sigdigits = 3)
-    println("3D=$(threeD), 1D=$(oneD)")
+    printstyled("Lower triangle only"; color = :green)
+    println("\nSpeedups: \n    3D=$threeD\n    1D=$oneD")
+    println(
+        """
+        Raw timings:
+            3D serial:   $(pretty_time(b_serial))
+            3D parallel: $(pretty_time(b_parallel))
+            1D serial:   $(pretty_time(b_s_1))
+            1D parallel: $(pretty_time(b_p_1))
+        """
+    )
+
+    # calculate the full matrix
+    b_serial_full   = @benchmark CrossRecurrenceMatrix($(tr), $(tr), 5.0; metric = Euclidean(), parallel = false)
+    b_parallel_full = @benchmark CrossRecurrenceMatrix($(tr), $(tr), 5.0; metric = Euclidean(), parallel = true)
+    b_s_1_full = @benchmark CrossRecurrenceMatrix($(x), $(x), 2.0; parallel = false)
+    b_p_1_full = @benchmark CrossRecurrenceMatrix($(x), $(x), 2.0; parallel = true)
+
+    # calculate full metrics
+    threeD = median(b_serial_full.times)/median(b_parallel_full.times)
+    oneD = median(b_s_1_full.times)/median(b_p_1_full.times)
+    threeD, oneD = round.((threeD, oneD); sigdigits = 3)
+    printstyled("Full matrix"; color = :green)
+    println("\nSpeedups: \n    3D=$threeD\n    1D=$oneD")
+    println(
+        """
+        Raw timings:
+            3D serial:   $(pretty_time(b_serial_full))
+            3D parallel: $(pretty_time(b_parallel_full))
+            1D serial:   $(pretty_time(b_s_1_full))
+            1D parallel: $(pretty_time(b_p_1_full))
+        """
+    )
+    println()
 end
--- a/src/matrices.jl
+++ b/src/matrices.jl
@@ -32,36 +32,39 @@ The list of strings available to define the metric are:
 * `"manhattan"`, `"cityblock"`, `"taxicab"` or `"min"` for the Manhattan or L1 norm
   (`Cityblock()` in `Distances`).
 """
-distancematrix(x, metric::Union{Metric,String}=DEFAULT_METRIC) = distancematrix(x, x, metric)
+distancematrix(x, metric::Union{Metric,String}=DEFAULT_METRIC, parallel = size(x)[1] > 500) = _distancematrix(x, getmetric(metric), Val(parallel))
 
 # For 1-dimensional arrays (vectors), the distance does not depend on the metric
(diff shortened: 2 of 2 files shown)


Datseris:
@asinghvi17 can you please paste the resultng benchmarks you have here as well?

asinghvi17:
Benchmark results:
```julia
I am using 4 threads.
For N = 101...
Lower triangle only
Speedups: 
    3D=1.17
    1D=1.18
Raw timings:
    3D serial:   220.835 μs (0.00% GC)
    3D parallel: 188.153 μs (0.00% GC)
    1D serial:   117.716 μs (0.00% GC)
    1D parallel: 99.657 μs (0.00% GC)

Full matrix
Speedups: 
    3D=1.22
    1D=1.12
Raw timings:
    3D serial:   405.159 μs (0.00% GC)
    3D parallel: 331.594 μs (0.00% GC)
    1D serial:   284.941 μs (0.00% GC)
    1D parallel: 254.926 μs (0.00% GC)


For N = 317...
Lower triangle only
Speedups: 
    3D=1.12
    1D=1.01
Raw timings:
    3D serial:   897.380 μs (0.00% GC)
    3D parallel: 803.080 μs (0.00% GC)
    1D serial:   651.474 μs (0.00% GC)
    1D parallel: 643.333 μs (0.00% GC)

Full matrix
Speedups: 
    3D=1.2
    1D=1.12
Raw timings:
    3D serial:   1.748 ms (0.00% GC)
    3D parallel: 1.455 ms (0.00% GC)
    1D serial:   1.321 ms (0.00% GC)
    1D parallel: 1.175 ms (0.00% GC)


For N = 1001...
Lower triangle only
Speedups: 
    3D=1.31
    1D=1.17
Raw timings:
    3D serial:   5.866 ms (0.00% GC)
    3D parallel: 4.486 ms (0.00% GC)
    1D serial:   5.409 ms (0.00% GC)
    1D parallel: 4.635 ms (0.00% GC)

Full matrix
Speedups: 
    3D=1.27
    1D=1.17
Raw timings:
    3D serial:   11.581 ms (0.00% GC)
    3D parallel: 9.091 ms (0.00% GC)
    1D serial:   10.671 ms (0.00% GC)
    1D parallel: 9.122 ms (0.00% GC)


For N = 3163...
Lower triangle only
Speedups: 
    3D=1.4
    1D=1.24
Raw timings:
    3D serial:   58.012 ms (0.00% GC)
    3D parallel: 41.296 ms (0.00% GC)
    1D serial:   61.319 ms (14.36% GC)
    1D parallel: 49.392 ms (0.00% GC)

Full matrix
Speedups: 
    3D=1.26
    1D=1.02
Raw timings:
    3D serial:   117.258 ms (9.34% GC)
    3D parallel: 93.262 ms (16.24% GC)
    1D serial:   114.398 ms (9.39% GC)
    1D parallel: 112.297 ms (17.50% GC)


For N = 10001...
Lower triangle only
Speedups: 
    3D=1.22
    1D=0.994
Raw timings:
    3D serial:   746.008 ms (7.89% GC)
    3D parallel: 610.439 ms (13.55% GC)
    1D serial:   729.746 ms (6.77% GC)
    1D parallel: 734.022 ms (14.27% GC)

Full matrix
Speedups: 
    3D=1.2
    1D=1.11
Raw timings:
    3D serial:   1.642 s (8.28% GC)
    3D parallel: 1.365 s (15.08% GC)
    1D serial:   1.758 s (5.23% GC)
    1D parallel: 1.581 s (14.81% GC)


For N = 31624...
Lower triangle only
Speedups: 
    3D=1.28
    1D=1.25
Raw timings:
    3D serial:   9.569 s (4.34% GC)
    3D parallel: 7.482 s (2.91% GC)
    1D serial:   11.350 s (5.22% GC)
    1D parallel: 9.070 s (7.27% GC)
```

asinghvi17:
OK, everything should work now.  Let me know if it doesn't, the options for the `Symmetric` constructor confused me a bit...

asinghvi17:
Failures on 1.0 are an upstream issue, specifically DiffEqBase.  I don't know what we can do about that...is the version restricted on our end?

heliosdrm:
Thank you so much for this great PR! Regarding the failed tests for Julia 1.0: the errors occur when trying to use DynamicalSystemsBase, but this is an "extra" package used only to create the test data sets, not a core dependency.

In my opinion, we should maintain a version compatible with 1.0, which is the LTS version of Julia. If new features like this were really incompatible with Julia 1.0, we could release a version 1.3.0 of RecurrenceAnalysis restricted to higher Julia versions, and keep maintenance of 1.2.x with LTS support, backporting bugfixes.

But perhaps this is still compatible with Julia 1.0, and ithe error of DynamicalSystemsBase in the tests could be worked around - maybe saving the data into text files, and making the tests read them with DelimitedFiles?

Datseris:
As you say, errors are related with DynSysBase and data generation, and I entirely agree that the packages used in the tests should not affect versioning here. Yet, I need to address the following:

> In my opinion, we should maintain a version compatible with 1.0, which is the LTS version of Julia. If new features like this were really incompatible with Julia 1.0, we could release a version 1.3.0 of RecurrenceAnalysis restricted to higher Julia versions, and keep maintenance of 1.2.x with LTS support, backporting bugfixes.

Can you please justify this more? Versions 1.x are a strict improvement of 1.0 and I find no argument (no argument whatsoever) to spend effort and developer's time supporting both "exclusively" 1.0 and 1.x... What you write here sounds like a simple thing to do, yet I stand to claim that maintaining 2 "release branches" of a package is a nightmare. Especially in our given case, where it is a package with specific and self contained functionality. Furthermore, Julia's support for parallelization seems to increase with every minor release. Given that these PRs are quite literally about parallelization and/or performance improvements, I think it is twice the argument to not stay in 1.0...

> But perhaps this is still compatible with Julia 1.0, and ithe error of DynamicalSystemsBase in the tests could be worked around - maybe saving the data into text files, and making the tests read them with DelimitedFiles?

This sounds a lot of work for me. And we wouldn't want to increase the file size of a repo. The test folder is downloaded as well when one installs any repo. Saving text files (especially large ones) is not a good idea.

---

I stand that we should just move forwards with the proper Julia version and not care. In order to be perfectly clear for the future, To be clear for the future, I just want to openly say that I will never be willing to spend any effort to support explicitly version 1.0. Of course, I will also not stop anyone from doing so, but I personally find it a waste of time.

asinghvi17:
I just made some changes locally to the distance matrix computation as well.  Will push when tested.

heliosdrm:
> Versions 1.x are a strict improvement of 1.0 and I find no argument (no argument whatsoever) to spend effort and developer's time supporting both "exclusively" 1.0 and 1.x...

My motivation for that statement is that there are users who may be bound to use Julia 1.0 until there is another LTS version - e.g. in very strict corporate settings. So I think that it is fair to let them have fixes to outstanding bugs that might be found in the future, e.g. wrong results or critical errors. Hopefully, such errors might never happen, and in that case maintenance of the version compatible with Julia 1.0 would be no effort at all.

So, what I propose is not working actively on two branches, but being positively reactive to issues that 1.0 users might submit, i.e. do not answer to them with "that version is not supported anymore", but pay attention to those issues or to PRs that might be merged into the last 1.0-compatible version.

> To be clear for the future, I just want to openly say that I will never be willing to spend any effort to support explicitly version 1.0. Of course, I will also not stop anyone from doing so, but I personally find it a waste of time.

This is very reasonable. I cannot promise anything, but if incoming issues related to 1.0 are not difficult to fix or backport, I volunteer to look after that.

Datseris:
Alright then, @heliosdrm I leave the merging of this PR to your hands then!

heliosdrm:
> > But perhaps this is still compatible with Julia 1.0, and ithe error of DynamicalSystemsBase in the tests could be worked around - maybe saving the data into text files, and making the tests read them with DelimitedFiles?
> 
> This sounds a lot of work for me. And we wouldn't want to increase the file size of a repo. The test folder is downloaded as well when one installs any repo. Saving text files (especially large ones) is not a good idea.

Agreed. On the other hand, I don't know why this is happening. The CI checks of the latest versions of DynamicalSystemsBase and DiffEqBase are from a few days ago, and they are alll passing. I would like to look into this, because I think we agree that compat bounds should not be changed only for the current failure, but at the same time it is not good to let it fail.

Datseris:
This problem is exclusively about DiffEqBase, all packages of DynamicalSystems.jl including DynamicalSystemsBase.jl have nothing to do with this problem whatsoever. 

This means, at least in my eyes, this can merged without a problem in any way. At some point DiffEqBase will fix its problem. From slack:

> Yes, it's known, there are two issues and a discussion about the problem in https://github.com/JuliaDiffEq/DiffEqBase.jl/issues/465 and https://github.com/JuliaDiffEq/DiffEqBase.jl/issues/462

heliosdrm:
One thing that confused me is that neither DynamicalSystemsBase (the direct dependency in RecurrenceAnalysis tests) or DiffEqBase (the actual offending package) showed CI errors, but now I see that the last check of DynamicalSystemsBase is older than the breaking changes of DiffEqBase (which has a very quick release cycle), and the newest versions of the latter do not support Julia 1.0.

Assuming that the developers of DiffEqBase (or the maintainers of the General Registry) will fix this soon, I'm fine with merging this PR, although it might be sensible to wait for the fix before tagging a new version. What do you think?

Datseris:
yes let's wait. https://github.com/JuliaDiffEq/DiffEqBase.jl/issues/465 is the issue to watch.

heliosdrm:
The issue in DeffEqBase.jl is fixed now, and tests pass again. 😃

PR #95, opened 2020-10-02
Transitivity and JRP product

Datseris:
Adds the `transitivity` as an RQA measure, and allows the convenience syntax `JointRecurrenceMatrix(R1, R2)` with R1,2 recurrence matrices.

@hkraemer can you provide some info I can put on the docstring of `transitivity` , like what does it mean or some paper that introduced the concept?

Also, should the Theiler window be taken into account here just like in `recurrencerate` ?


heliosdrm:
About the Travis failures:
 - `distance_matrix.jl` has not been `include`d. 
 - There are compatibility issues with Julia 1.0 again. That restricts `DiffEqBase` to 6.20 or lower. On the other hand, something else (I'm not able to find it out) is requiring `RecipesBase` to be 1.1, and this is not compatible for `DiffEqBase` lower than `6.26`.

Datseris:
The rest of the packages that compose DynamicalSystems.jl have already moved to Julia 1.5 and take advantage of the nice new syntax with keyword arguments, and also don't have to cope so much with packages that have problem getting the dependencies right. Do you mind if we move this one to 1.5 as well? Users using this package through DynamicalSystems.jl will not be able to get updates for julia 1.0 anyways. Also, in this move to 1.5 we have dropped all plotting related dependencies; RecipesBase won't be asked anymore.

Datseris:
@hkraemer I've just pushed a sketch for the FAN method you propose. Feel free to modify accordingly until it "makes sense" and then I'll take care of smoothening out the API.

Datseris:
@hkraemer I've removed the fan code from this PR. I think for organization purposes it is better if FAN is added in its own individual PR. For reference, the code was:
```julia
function recurrence_matrix_fan(..., k)
    d = distancematrix(x, y)
    for i in 1:size(d, 1)
        ε = some how transform percentange into ε
        d .< ε
        ...
    end
end
```

heliosdrm:
> The rest of the packages that compose DynamicalSystems.jl have already moved to Julia 1.5 and take advantage of the nice new syntax with keyword arguments, and also don't have to cope so much with packages that have problem getting the dependencies right. Do you mind if we move this one to 1.5 as well? Users using this package through DynamicalSystems.jl will not be able to get updates for julia 1.0 anyways. Also, in this move to 1.5 we have dropped all plotting related dependencies; RecipesBase won't be asked anymore.

Fine for me. I was wary of dropping compatibility with the Julia LTS if it could be kept without much pain, but since the promised next LTS version (1.6) is probably around the corner, I think it's no big deal now.

Datseris:
@heliosdrm tests fail because of this new function I've made, called `transitivity`, which I also added in the full RQA stack of functions.

What this showed me is that the CRP you create in the file `smallmatrix.jl` is in fact not a square matrix, it is (10,11). I am wondering, does non-square recurrence plots make sense...? If they do then it is fine, I will simply not compute `transiitivity` for non-square matrices.

heliosdrm:
> What this showed me is that the CRP you create in the file `smallmatrix.jl` is in fact not a square matrix, it is (10,11). I am wondering, does non-square recurrence plots make sense...? If they do then it is fine, I will simply not compute `transiitivity` for non-square matrices.

Yes, joint recurrence plots can be non-square, e.g. if the two series represented in the plot have different lengths. So, catching that error in `smallmatrix.jl` was good!

pucicu:
cross recurrence plots can be non-square, because the vectors x and y can have different lengths. joint cross recurrence are square because they are element-wise products of the recurrence matrices of x and y.

Datseris:
thanks a lot @hkraemer !

PR #97, opened 2020-10-12
Fan threshold method

hkraemer:
Hi @heliosdrm and @Datseris ,
I implemented the fixed amount of neighbors threshold selection method. As with the `fixedrate` this will lead to a RP with a fixed global recurrence rate. However, here we vary the threshold for each point on the trajectory. This procedure corresponds to a fixed local recurrence rate. This method particulary performs well on data not sampled well, since it can capture recurrence structures adaptively.
I tried to incorporate the changes within your structure and simply added the methods, allowing for dispatching on it, when the recurrence threshold is a `Vector`. I also included some tests in `test/dynamicalsytems.jl`. Here, I changed the time series to be tested a little bit. I added white noise of negligible amplitude to the time series. This won't change anything in the tests, but for the fan-method this is neccessary, when it comes to perfectly period motion. In this case, and in case of very small time series length, as used here in the tests, the local recurrence threshold can not be estimated accurately, since there are so many distances == 0. 
Some style argument: I had to expand `resolve_scale`-function and binded the `kwargs` in a very "ugly manner". - I am sure you find a cleaner and nicer solution?!
Cheers


src/matrices.jl
```diff
@@ -351,23 +165,60 @@ function JointRecurrenceMatrix(x, y, ε; kwargs...)
...
-    if fixedrate
+    if fixedrate && fan == false
         sfun = (m) -> quantile(vec(m), ε)
         return resolve_scale(Base.front(args)..., 1.0; scale=sfun, fixedrate=false)
+    elseif fixedrate && fan || fixedrate == false && fan
+        try
+            return get_fan_threshold(x, y, metric, ε)
+        catch
+            return get_fan_threshold(x, x, metric, ε)
+        end
```
Datseris, line 199: Try-catch statements should not be used in performance-related code. What do you try to do here?

src/matrices.jl
```diff
@@ -351,23 +165,60 @@ function JointRecurrenceMatrix(x, y, ε; kwargs...)
...
+        metric = args[3]
+    else
+        metric = args[2]
+    end
     # Check fixed recurrence rate - ε must be within (0, 1)
-    if fixedrate
+    if fixedrate && fan == false
         sfun = (m) -> quantile(vec(m), ε)
         return resolve_scale(Base.front(args)..., 1.0; scale=sfun, fixedrate=false)
+    elseif fixedrate && fan || fixedrate == false && fan
```
Datseris, line 194: This is a bit hard to read : `fixedrate && fan || fixedrate == false && fan` Can you please add parenthesis appropriately?

src/matrices.jl
hkraemer: Good point, I get rid of it, by binding `y` right on top of the function.

src/matrices.jl
hkraemer: Done.

Datseris:
@heliosdrm could you please become the reviewer of this PR? It is well beyond my bandwidth at the moment. If neither of us can be a reviewer, I think I'll leave it to the judgement of @hkraemer to merge when he feels it is ready. He is also an expert in Recurrence Plots.

heliosdrm:
> @heliosdrm could you please become the reviewer of this PR? 

I'm happy to review it... if it is possible to hold it for two or three weeks. I've got some deadlines in October and very few time to spend right now. If it's too much time, I agree with leaving it to @hkraemer

src/matrices.jl
```diff
@@ -351,23 +165,57 @@ function JointRecurrenceMatrix(x, y, ε; kwargs...)
...
+        metric = args[2]
+        y = x
+    end
     # Check fixed recurrence rate - ε must be within (0, 1)
-    if fixedrate
+    if fixedrate && (fan == false)
         sfun = (m) -> quantile(vec(m), ε)
         return resolve_scale(Base.front(args)..., 1.0; scale=sfun, fixedrate=false)
+    elseif (fixedrate && fan) || (fixedrate == false && fan)
+        return get_fan_threshold(x, y, metric, ε)
```
heliosdrm, line 196: I'm not happy with this: if the keyword argument `fan` is false, `resolve_scale` will return a number, but if it is `true`, it will return a vector, thus making the whole thing type-unstable.

src/matrices.jl
```diff
@@ -351,23 +165,57 @@ function JointRecurrenceMatrix(x, y, ε; kwargs...)
...
+end
 
 ################################################################################
 # Scaling / fixed rate
 ################################################################################
 # here args... is (x, y, metric, ε) or just (x, metric, ε)
-function resolve_scale(args...; scale=1, fixedrate=false)
+function resolve_scale(args...; scale=1, fixedrate=false, fan=false)
     ε = args[end]
+    x = args[1]
```
heliosdrm, line 183: I'd suggest that instead of defining the arguments for `get_fan_threshold` based on the length of `args`, the function `get_fan_threshold` could be redefined with different methods (for three or four arguments), such that `args` can be passed directly to it.

heliosdrm (changes requested):
Sorry for taking so long to review this. I see two major things to improve:

Most important: `resolve_scale` becomes type-unstable, and this should be avoided. I see two ways of working around this:
1. Modify the other `return` lines of `resolve_scale`, such that it always returns a vector - with a single element where it is meant to be a scalar. All the other functions that take the value returned by `resolve_scale`, i.e. `recurrence_matrix`, etc. should be modified, such that their behavior depends on the length of that argument. (See below about this.)
2. Turn `resolve_scale` into two different functions, for FAN (returning a vector) and non-FAN (returning a scalar). The functions that now call `resolve_scale` already have some burden of type ambiguity, because they define `Val(parallel)`, whose type is unpredictable; so I think that it would be bearable for them to call either version of `resolve_scale`.

The other issue that I see is that there is a lot of redundancy in the new code, with FAN versions of many functions that only differ in a couple of lines:
* An `@assert` line added to check that the threshold is as long as the input.
* The line where the distance between points is compared with `ε`, which is changed to `ε[i]`.

The code would be much more maintainable if that redundancy is removed. E.g.:
* Removing or modifying the `@assert` line, such that it passes if `ε` is a scalar.
* Changing `ε[i]` by `ε[min(i, n)]`, where  `n` is previously defined as the length of `ε`. For scalars this is `ε[1]` which is the same as `ε`.
This would also work with the first suggested solution to the previous issue, i.e. changing scalar thresholds by single-valued vectors.

src/matrices.jl
```diff
@@ -351,23 +165,57 @@ function JointRecurrenceMatrix(x, y, ε; kwargs...)
...
+    else
+        metric = args[2]
+        y = x
+    end
     # Check fixed recurrence rate - ε must be within (0, 1)
-    if fixedrate
+    if fixedrate && (fan == false)
         sfun = (m) -> quantile(vec(m), ε)
         return resolve_scale(Base.front(args)..., 1.0; scale=sfun, fixedrate=false)
+    elseif (fixedrate && fan) || (fixedrate == false && fan)
```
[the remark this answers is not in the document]
hkraemer: true that. :) changed it

src/matrices.jl
hkraemer: yepp, you are absolutely right. I followed your 1st suggestion.

src/matrices.jl
hkraemer: yepp, also right. I couldn't simply help myself in the first turn, because I did not (and I think still do not) know much about string interning.

hkraemer:
Hey @heliosdrm ,
thank you for your review and your valuable suggestions. I followed your 1st suggestion, thus, `resolve_scale` is now type-stable. Regarding the redundancy in the `recurrence_matrix`-methods. I reduced it, but the problem is, that I have to make a case distinction at some point, due to the non-symmetry of the recurrence matrix in case of FAN-threshold selection. Anyway, due to the other changes I made (which you suggested), it should be easier to read/maintain. 
Cheers

heliosdrm:
Hi @hkramer, you're right. I had overlooked the effect of the FAN method on the symmetry of recurrence matrices. This has made me discover the issue mentioned in #100. I'm considering other modifications by now, to make potential further changes cleaner. I'll write them during this week and submit them to this same pull request, if you are ok with that.

heliosdrm:
Part of the changes to this PR I'm thinking on consist of allowing the introduction of `fan::Val` as a positional argument, not a keyword argument, in the constructor of `RecurrenceMatrix`, etc., such that if we make it a parametric type with potentially `Symmetric` data, its type will be predictable from the arguments passed on to the constructor. @Datseris : do you think this is suitable?

src/matrices.jl
```diff
@@ -296,10 +110,11 @@ recurrence quantifications", in: Webber, C.L. & N. Marwan (eds.), *Recurrence
...
-    parallel::Bool = length(x) > 500 && Threads.nthreads() > 1, kwargs...)
+    parallel::Bool = length(x) > 500 && Threads.nthreads() > 1, fan::Bool = false,
+    kwargs...)
     m = getmetric(metric)
-    s = resolve_scale(x, m, ε; kwargs...)
-    m = recurrence_matrix(x, m, s, Val(parallel))
+    s = resolve_scale(x, m, ε; fan, kwargs...)
+    m = recurrence_matrix(x, m, s, Val(parallel), Val(fan))
     return RecurrenceMatrix(m)
 end
```
Datseris, line 120: @heliosdrm does your question target this function, where you said that `val` should be positional instead of keyword? If so, it makes no difference, since it is already positional in the `recurrence_matrix` call. Besides, you can never get rid of the type istability of going from `fan` to `Val(fan)` in every single function call. One function call will always remain type unstable, and it is best if this call is the highest level one.

Datseris:
May I suggest that instead of messing with all this `Val` stuff, which is not recommended in general Julia programming, you create 2 singleton types:
```julia
struct FAN end
struct Rate end # better name here
```
and the user can either give `FAN()` or `Rate()` arguments to the creator of `RecurrenceMatrix`? This is perfectly type stable, allows dispatching on whether the matrix has been created with either of the two approaches, and the type information can also be included in the recurrence matrix as a type parameter.

src/matrices.jl
heliosdrm: I mean in the constructor of `RecurrenceMatrix` itself. Now it is:

`RecurrenceMatrix(x, ε; kwargs...)`

There could be another method `RecurrenceMatrix(x, ε, fan::Val; kwargs...)`, such that if `fan isa `Val{false}` the result will contain a `Symmetric` sparse matrix, and if `fan isa `Val{ŧrue}` it will be a non-symmetric one. A bit less friendly than using `fan::Bool` in `kwargs`, as currently proposed. But we could have that method as a function barrier, available
together with the friendlier but type-unstable one.

heliosdrm: I like the idea of using the singleton types instead of `Val` types.

heliosdrm:
Shall we close this one in favor or #102?

PR #101, opened 2020-11-24
Fan

hkraemer:
re-organized PR #97 , so here only the changes really affecting the FAN-threshold method


hkraemer:
hopefully that is what you wanted, @Datseris  and @heliosdrm ?!

hkraemer:
okay, I've seen that there is the "old stuff" included... I closed it

Datseris:
lol git can be a pain in the ass sometimes, I've felt this way too many times :D

Datseris:
@hkraemer there are some commits from me here that say `reorg code`, so maybe I also did that during the PR? Then we have to merge these into master, there is no way to get rid of them.

hkraemer:
yepp, I think so :/

hkraemer:
and all the transitivity stuff... but as @heliosdrm said, basically we only need the commits from da507d6af0febf283c1214fc023720bc1b2b84a4 onwards.

Datseris:
Um, but the transitivity stuff are merged and tagged in version 1.3.0. So I think we're good?

Issue #100, opened 2020-11-23
`RecurrenceMatrix` does not take advantage of symmetry  


heliosdrm:
The data of `RecurrenceMatrix` is of type `SparseMatrixCSC{Bool,Int}`. This means that, although the constructor passes a symmetric matrix with only the entries of the upper triangle, the underlying data in the resulting `RecurrenceMatrix` makes a new "regular" sparse matrix duplicating them on the lower triangle.

As it is in version 1.3, this could be changed without modifications in the API if the data field of `RecurrenceMatrix` is changed to `Symmetric{Float64,SparseMatrixCSC{Bool,Int64}}`. However, with the FAN threshold method (#97) recurrence matrices can also be non-symmetric, so:
* Right now it is better to keep it as it is, to allow the introduction of the FAN threshold method without changing the API.
* For the next major version we could consider converting `RecurrenceMatrix` (and maybe the other types?) into parametric types, to take advantage of potential symmetry.

Datseris:
> For the next major version we could consider converting `RecurrenceMatrix` (and maybe the other types?) into parametric types, to take advantage of potential symmetry.

Changing the type parameterization of `RecurrenceMatrix` is in no way breaking, we could do it as soon as there is a PR that implements it. `RecurrenceMatrix` can be parameterized on the type of its `data` field.

heliosdrm:
Good, I'll write another PR in parallel with #97, taking care of making them mergeable without too many conflicts (I hope).

Datseris:
I have *yet* another suggestion @hkraemer and @heliosdrm . Pr #97 is **massive**. Its git difference is very large, and difficult to follow. Perhaps we can clean things up by :  

1. Removing the *genuinely new* FAN-related code from #97.
1. Merging changes.
1. Opening a *new* pr with the actual FAN changes.
1. In the *same* PR, add the type information additions which allow us to know whether a RM is symmetric based on its type.

I know this sounds like tedious work, but there is actually only one reason I suggest it: it will make mistakes/errors/bad design decisions impossible to hide.

heliosdrm:
No need to merge non-FAN related changes from #97. It took me a while to catch it, but the 13 first commits of that branch were already squashed-merged into 1.3.0. Only its five last commits have relevant changes, and they are limited to `matrices.jl` and a few new tests.

[...]

PR #102, opened 2020-11-24
FAN method

heliosdrm:
New attempt of re-organizing #97


Datseris:
Okay great that everything is resolved. Then I take it that we can proceed with adding the new type-information here!

heliosdrm:
Regarding the symmetry issue discussed in #100:

Currently `RecurrenceMatrix` is symmetric by definition (`JointRecurrenceMatrix` is not, but it should be, too). This allows some optimizations; e.g. `issymmetric` always returns `true` for that type, without looking into its contents. This is relevant for the calculation of diagonal structures, which uses `issymmetric` to check if they can be obtained by only traversing the upper triangle.

With the FAN method, recurrence matrices may be non-symmetric. So, in order to keep the advantages of the current approach, the result of using FAN should be of a different type. These are the alternatives I'm thinking of:
* Create a new type, e.g. `FANRecurrenceMatrix` special for that method.
* Make `RecurrenceMatrix`, etc. parametric types, such that the parameter can be used for symmetry checks. To use the FAN method, an argument of special type could be passed to the constructor, to deliver the correct result -- e.g. the singleton type `FAN` suggested in https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/pull/97#issuecomment-732459048

The first option is simpler, and the second smarter. Opinions? Alternatives?

Datseris:
The first option is, in my eyes, more complicated than the second. Additional constructors would have to be re-written for all the `FANRecurrenceMatrix` instances, and the API is increased by having more methods and types. For me the best solution here is a singleton type as type parameter of `RecurrenceMatrix`. It has the additional benefit of only increasing method complexity when necessary. All methods that don't care about symmetry dispatch on `RecurrenceMatrix` without a problem, while methods that do care, can dispatch on the type parameterization.

heliosdrm:
Do you mean this?:

```julia
RecurrenceMatrix(args...) # Fixed threshold, as it is now
RecurrenceMatrix{FAN}(args...) # Using FAN method
```

The second one would obviously return an object of type `RecurrenceMatrix{FAN}`. The first one would return a `RecurrenceMatrix{???}` (the question about how to call the alternative type parameter is still open).

Datseris:
Well you can expose the methods `RecurrenceMatrix{M1}` and `RecurrenceMatrix{M2}` but you can also just let it at a higher level like 
```
function RecurrenceMatrix(...; fan = true)

if fan
do_fan_stuff
return RecurrenceMatrix{FAN}(...)
else
do_non_fan
return RecurrenceMatrix{NonFAN}
end
end
```

p.s. what does FAN stand for? Fixed Amount of Neighbors?

hkraemer:
yepp, fixed amount of neighbours

heliosdrm:
About the parameter to be used when the FAN method is *not* applied: if there is nothing obvious, we could use `Nothing` indeed. 😃  I mean, that e.g. `RecurrenceMatrix(...)` would return a `RecurrenceMatrix{Nothing}` object.

By the way, I think that using a type parameter to identify special ways of computing recurrence matrices is a great idea, specifically if the results have special properties and their analysis could be optimized with special methods. It would be very straightforward to write such special methods dispatching on `AbstractRecurrenceMatrix{FOO}`, for matrices made with a new `FOO` method, etc. Moreover, the parameter itself serves as "meta-data" that clearly identifies how the matrix was made, if that's relevant. (Of course, this feature should not be abused, to avoid overpopulating the method tables.)

Datseris:
I would still vote to not go with `Nothing` for the other option. What we are talking about here is the classical case of fixed range neighbors versus fixed number (also called fixed mass) neighbors. It is the same as doing a `knn` search or an `inrange` search. Thus, I propose calling the two names as `FixedAmount` or `FixedRange`.

heliosdrm:
> I propose calling the two names as `FixedAmount` or `FixedRange`

Perhaps with an alias `const FAN = FixedAmount`, since the acronym is quite known in the field?

heliosdrm:
I was thinking on what would be the best way of tweaking `recurrencematrix`, in order to reduce redundancy of code, and at the same time be as optimal as possible.

In 5bb1ec3, after my comments on #97 @hkraemer made `ε` be always a vector - filled with the values of the scaled threshold when previously it was a scalar. However, this allocates unnecessary values. I think it's better to keep it as a scalar, and tweak lines like this:
https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/pull/102/files#diff-a8a8c64dc575dda34081050aed4efc90ec048136b4385c3fa7f1e05b6ae8d80cR312-R312
to pick `ε[1]` instead of `ε[i]` during the iterations when `ε isa Real`.

However, that would mean an additional (albeit cheap) operation at every iteration, like indexing on `min(i, length(ε))` instead of `i`. So, inspired by [FillArrays.jl](https://github.com/JuliaArrays/FillArrays.jl), I thought on a possible (hackish) alternative:

When `ε` is meant to be a scalar, wrap it into a custom type like this:

```julia
struct Always{T}
     data::T
end
Base.getindex(x::Always, i) = x.data
```

Thus, if `ε = Always(value)`, then we could call `ε[i]` with whatever index `i`, and always get `value`.

But maybe this is too far-fetched, and potentially misleading for future revisions of the code... So I'm not sure. What do you think?

Datseris:
> But maybe this is too far-fetched, and potentially misleading for future revisions of the code... So I'm not sure. What do you think?

This is way too far fetched. We are talking about optimizations that have miniscule, almost unmeasurable performance impact.

The cost of making a small vector is nanoseconds:

```
julia> using BenchmarkTools
[ Info: Precompiling BenchmarkTools [6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf]

julia> @btime [1, 2, 3];
  18.800 ns (1 allocation: 112 bytes)
```
My two cents is that we should go for the clearest code here, the code that reduces duplication and is easiest to read.

src/matrices.jl
```diff
@@ -293,13 +340,14 @@ function recurrence_matrix(x::AbstractVector, y::AbstractVector, metric, ε, par
...
-function recurrence_matrix(x::AbstractVector, metric::Metric, ε::Real, parallel::Val{false})
+function recurrence_matrix(x::AbstractVector, metric::Metric, ε::Vector, parallel::Val{false}, fan::Val{false})
+    @assert length(ε) == length(x)
     rowvals = Vector{Int}()
     colvals = Vector{Int}()
     for j in 1:length(x)
         nzcol = 0
         for i in 1:j
-            if @inbounds abs(x[i] - x[j]) ≤ ε
+            if @inbounds evaluate(metric, x[i], x[j]) ≤ ε[i]
```
heliosdrm, line 350: I do not understand the change in line 350: If `x` is a vector, all metrics should give the same result as `abs`, isn't it? Perhaps it's an accidental copy-and-paste from another method. It also happens in the next method, when the FAN method is applied.

Datseris: yeap you are right, these two lines coincide when `x` is a vector

heliosdrm:
In 450b77f I have removed a lot of redundancy by making the constructor `RecurrenceMatrix{FixedAmount}` call `recurrence_matrix(x, x, ...)`. That way, there is no need of special methods of `recurrence_matrix` with a single input that may create non-symmetric results.

In the tests, there are some warnings of transitivity not being applicable to non-square matrices. Maybe `@test_warn` should be used in those cases?

Datseris:
BTW guys, have you ever considered using KDTrees to construct a recurrence matrix? It might be much faster. 

We recently came across issue when computing correlation sums, which in their core are very similar to calculating a recurrence matrix.

In principle we could be making a KDTree and then iterating all points and getting the indices of neighbors that are `inrange(ε)`, and put a 1 in these indices. Same story hapens for the FAN method but instead of `inrange` we can call `knn`.

heliosdrm:
When `fixedrate==true`, the threshold is scaled to ensure a fixed recurrence rate *excluding the LOI* if you are calculating a `RecurrenceMatrix`. But as it is now, with FAN the whole column is considered. We should change that to be consistent, don't you think?

heliosdrm:
> BTW guys, have you ever considered using KDTrees to construct a recurrence matrix? It might be much faster.

So far I have not used KDTrees. It may be a good idea, but don't know how to do it. May be worth to open an issue for that.

Datseris:
I know how and I can try. But I'd prefer to do this after we've merged and stabilized this PR.

Datseris:
> When fixedrate==true, the threshold is scaled to ensure a fixed recurrence rate excluding the LOI if you are calculating a RecurrenceMatrix. But as it is now, with FAN the whole column is considered. We should change that to be consistent, don't you think?

I'm a bit confused. Isn't "fixed rate" and "FAN" the same thing? If you have a fixed recurrence rate, isn't that equivalent with having a fixed amount of neighors for each point?

Datseris:
BTW, just as a heads up @heliosdrm : the whole DynamicalSystems.jl ecosystem is preparing for a 2.0 release, to allow some breaking changes. I have been putting deprecation warnings throughout the library so far. If there is something here that we want to change, but are reluctant to because it will breaking, this will be a good point in time to do it.

heliosdrm:
> I'm a bit confused. Isn't "fixed rate" and "FAN" the same thing?

"fixed rate" means setting the threshold in order to have a fixed (pre-defined) recurrence rate for the whole matrix, but each point can still have a different number of neighbors.

What I was commenting is that, following common practice, when there is a single input, that recurrence rate excludes considering every point a neighbor of itself. But this doesn't happen with FAN right now, so I was proposing to change that.

heliosdrm:
@hkraemer : `get_fan_threshold` requires both time series to be the same, but that does not allow to create cross recurrence matrices for time series of different length. Is there are compelling reason for that restriction?

Datseris:
I agree that the number the decides FAN does *not* include the self.

heliosdrm:
I have nothing else to revise of the PR so far. If @Datseris and @hkraemer are happy with the changes, I think this could be merged.

pucicu:
we had a diploma project on this. I’m not sure whether this would have covered your idea, but the student was investigating different clustering and approximation approaches (I think it also included KD trees) and the effect on the accuracy. unfortunately, the outcome was not promising, i.e., there was either not much benefit from speed or the accuracy was too bad.

David Salomon: Bestimmung der Reccurence Matrix für die Reccurence Quantification Analyse mittels Approximate Nearest Neighbour Search
Diploma thesis, Humboldt Univ Berlin, 2016

Datseris (approved):

hkraemer:
Hey, 
there is som inconsitency in the code: When parsing non-`Dataset`-type-Vectors into `RecurrenceMatrix` or `CrossRecurrenceMatrix` an error is thrown: 

- In the first case `::Array{Float64,2}` is not suported, one would need to parse in `Dataset(data)`.  Similar to the [approach used in case of constructing a `CrossRecurrenceMatrix` ](https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/be0443141da80ed703a2108c7a2e51db83a40764/src/matrices.jl#L319-L323) one could simply transform `AbstractMatrix`-inputs into `Dataset`-format. This way, one would also [get rid of an additional method](https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/be0443141da80ed703a2108c7a2e51db83a40764/src/matrices.jl#L467-L493).

- In the second case, an `assertion error` is thrown, based on [this](https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/be0443141da80ed703a2108c7a2e51db83a40764/src/matrices.jl#L352) and [this](https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/be0443141da80ed703a2108c7a2e51db83a40764/src/matrices.jl#L441) line of code, when the input `data` is a multidimensional trajectory. The problem here is that `length(data)` is not giving the trajectory length, when data is not of type `Dataset`, but a normal `Array`. One could simply use `size(data,1)` in this case. Regarding [this code ](https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/be0443141da80ed703a2108c7a2e51db83a40764/src/matrices.jl#L319-L323), I do not understand why the mentioned methods are necessary, nor why they actually get called? This should all be transformed into `Dataset`-format, no?

Shall we fix these things in this PR or in another one?

Datseris:
Yes, here is better. My take is that there is no reason to support `::Array{Float64,2}` (which is _multidimensional_ data), because the expected format for this data type is `Dataset`. `Array{T, 1}` is a different story, because this is a _timeseries_ and should be supported everywhere, because the expected format for timeseries is `Array{T, 1}`.

hkraemer:
But why don't you think we should not suppert `::Array{Float64,2}`? From a user-perspective it does make sense to allow it and internally transform it into `Dataset`-format. It is not much effort...

heliosdrm:
The problem is actually in the line of `get_fan_threshold` which says:

```julia
    fan_threshold = zeros(length(y))
```

This makes the vector of thresholds be longer than expected when the input is a matrix.

Then there is a method `recurrence_matrix(x::AbstractMatrix, y::AbstractMatrix, metric::Metric, ε, parallel::Val)` that converts the inputs to `Dataset`, so that `length(y)` etc. becomes of the right length (but inconsistent with that of `ε`).

The fix is easy: change the previous line to to `fan_threshold = zeros(size(y,1))`, and also add the method of `recurrence_matrix` for a single `AbstractMatrix` input. But according to @Datseris 's comments here and in #106, maybe we should rather *remove* the method for two `AbstractMatrix` inputs, and deprecate `AbstractRecurrenceMatrix{T}(::AbstractMatrix, args...) where T`. Right?

Datseris:
Well I guess this will be a long rant, but its important to write, because the goal of Julia and DynamicalSystems.jl isn't to baby-pamper expectations from other languages, but to exceed in quality all of its predecessors.

> It is not much effort...

Not if you only do it for one function, that's true. DynamicalSystems.jl has several dozen functions that operate on multidimensional input. For me always adding an additional clause to every function that will join the library is unreasonable. Of course feel free to add clauses wherever you want, but I doubt you will be able to support this train of thought in the long run. Bugs will creep up to you (just like now) because of the very simple reason: you take a change that should be done in one point (the user's input, or better yet the user's brain) and spread it over 100s of points (every function that accepts `Dataset` as input). So you literally multiply bug possibilities by 100.

There is something else in the air I should address about `Dataset` vs `Matrix. While in languages like Matlab or Python representing multidimensional data as a `Matrix` makes sense, in Julia it simply doesn't. You _can_ do it, but you _shouldn't_. So the idea of always trying to allow `Matrix` input is for me flawed, because it **literally teaches wrong usage**. This is doubly true given that there is not real counter argument for allowing `Matrix`: the documentation is clear, multi dimensional data are represented as `Dataset`, that's the end.

Here is another example: using strings or symbols to represent metrics, like e.g. `:euclidean` or `"euclidean"`. In the Julia language it has been widely accepted, and clear, that metrics are subtypes of `Metric` from the package Distances.jl. Users specify metrics as `Euclidean()`, not as `"euclidean"`. This is true not for 1 or 2, but for 100s of Julia packages. Yet in RecurrenceAnalysis.jl we still use `"euclidean"`. Now, you could add additional clauses and boilerplate code in all the libraries so that they translate string to Metric. This is what we do here for example. Do you think it is a good idea long term? I absolutely don't think so. Instead using input arguments we just add more and more code without any real benefit. 

Like, I don't get it. What's the big deal in typing `Dataset(your_matrix)`? Or the big deal in using `Euclidean()` instead of `"euclidean"`? It's better to learn to use a library and its API properly than to add boilerplate code to support expectations from other languages. That's my take on the topic.

pucicu:
Since I’m the reason for this discussion, I just would like to add some thoughts.

First, every time I try to use Julia (and Python too) I’m struggling with errors because functions do not accept my input arguments. For the same thing, there is a number of different variants to define it. In MATLAB, I can use vectors and matrices with the same functions. The rate to get an error in MATLAB is for me smaller than 10%, for functions or toolboxes that I have never used before. In Python and Julia, even for the simplest stuff I would like to do, the rate to get errors is 90%, because the functions do not support this or that data type or the different format of lists, arrays, matrices, datatables, etc. is expected.

OK, you are right, there are good reasons for data types. But the thrown error messages are usually several lines long and not easy to understand. Compare it with MATLAB – it is just one or two lines with the most precise statement that even helps to solve problem.

Coming back to current problem: I completely understand George. But I would expect that “Dataset” should be something more standard. I was looking for it with my favorite search web engine and have not found anything. I got hints on Julia packages DataFrame etc. but nothing that helped me to solve the problem. One solution would be to have clear advice in the thrown error message about the Dataset type. I was also confused that it was possible to use 1D-array. Why should it then not also be a Dataset?

Would it be possible to promote the Dataset type for use other Julia packages to have it like a standard for similar use cases? One problem I see with Julia and Python is that because so many different groups develop independent packages that might be used together in one piece of code is that similar data types or similar functions with different input data types compete, even within the same piece of code. In MATLAB this (almost) cannot happen because it is simple by design.

I don’t think that we discuss here to "baby-pamper expectations from other languages”. But it is more a question what a user would expect how it would work. If 90% of the users have difficulties in a first try to apply a function correctly, because the input should be of a certain format or data type, then perhaps it is better to re-design the code in a way to have it less error prone. (excuse me for this speculated number, it should only illustrate the worst case – I don’t mean a specific package)


In MATLAB, I like very much that I can simply call “help function” and I get a clear explanation what the function is doing and, this is the best!, some examples how to use the function. Instead in Python and Julia, documentation is usually full of data types clattering the essential information. However, this is indeed necessary because the functioning is so much depending on the selection of the correct data types. But this makes it simply very difficult to find the correct solution. DynamicalSystem.jl is here a great exception, because it has already a very good documentation and many examples (perhaps it could still be improved at some places, in particular in the function documentation that is available in the Juno documentation window). I would like to have a similar good documentation in general and also with many examples and how to use them in different situations.

Sorry for this lengthy explanation of my thoughts. I don’t expect to change anything, because Julia and Python cannot change this by definition and this is ok for me. But at least the documentation and the way how error messages are designed could be improved.

And to make it clear: I really appreciate the efforts you are doing! I am really interested to use Julia more frequently (and your package is helping here a lot!). But up to now, every use of Julia (or Python) ends with a lot of frustration. Perhaps it would help to have something like a table with “MATLAB example” – “How to do it in Julia” (I know that there is such an attempt somewhere but it is still in its infancy and it does not explain well how to do it).

hkraemer:
I am okay with all opinions. @Datseris 's statement boils it down to the essentials from a developer perspective, @pucicu 's statment does the same from a user point of view. I am actually somewhere in between and accept anything :) Highlighting the `Dataset`-Type maybe even more in the docs and also in clearer error messages sounds reasonable, because unlike the `Metric`-type (very good example btw. and I'd also vote for making this compatible in `RecurrenceAnalysis.jl`), `Dataset`-type is only used in `DynamicalSystems.jl`. Up to you. If there is anything to do, I am willing to help.
Back to the little problem: @heliosdrm is right with respect to `fan_threshold = zeros(size(y,1))` in `get_fan_threshold`, I think. The other stuff with the Matrices you have to discuss.

heliosdrm:
The problem related to matrix inputs should be solved now, with the addition of (deprecated) methods. Is there anything else to be fixed?

heliosdrm:
Can we merge this? I see no further blocking issue.

Datseris:
During the creation of the Neighborhood.jl package, a lot of discussion was put into choosing names for "FixedRange, FixedAmount,". Althought I'm not entirely satisfied with the outcome, the community agreed with the following two:

`NeighborNumber(k::Int), WithinRange(ε::Real)`

Over the past couple of weeks I've been moving more and more code of DynamicalSystems.jl to Neighborhood.jl. Now these two names `NeighborNumber, WithinRange` are a proper part of the ecosystem. I think we should incorporate them here as well instead of creating new names to dispatch on and introduce more terminology for what is inprinciple exactly the same thing. Original these terms are known in nonlinear dynamics as "FixedMass" and "FixedRange" neighborhood (which I think is by far the worst).

So, to make a long story short, instead of defining new terms `FixedRange, FixedAmount,` please use the existing ones. They are already in DelayEmbeddings and you can use them directly, as DelayEmbeddings exports them.

Datseris:
(as we speak I'm updating the "neighborhood" documentation page of DynamicalSystems.jl to reflect the new approach of using Neighborhood.jl, as I've just finished porting all code there)

Datseris:
@pucicu I've finally found some time to reply to your message. This conversation is getting well off topic, not only for this PR, but for DynamicalSystems.jl in general, so I'll keep it short. I think it is to your benefit, and to the benefit of the Julia community, if you start such discussions in the official Discourse forum here: https://discourse.julialang.org/ and provide insights of what a MATLAB user is missing when using Julia. So please, if you want to reply to this, do it on discourse instead of here.

From reading your message in detail, I think that you will start being much happier with how things are once you accept the following points:

* Julia and all of its packages are a community driven project where almost everyone contributes almost exclusively on their free time. As a hobby. So, consider: would I want to spent my hobby time to allow "matrix" to be used in RQA (without any actual functional improvement to the library, since you can do `Dataset(matrix)`), instead of implement some cool new feature, or fixing a crucial bug? I'd say nope.
* Julia's core programming paradigm is Multiple Dispatch, very much different from how MATLAB works.
* Julia is much younger than MATLAB and didn't have the same amount of time to fix bugs or to fine tune documentation aspects.

You're right, many things in Julia aren't as polished as MATLAB. And we are all working hard to improve things. In the past two weeks I have been writing documentation and polishing interfaces for DynamicalSystems.jl almost every night. But on the other hand, these are the things that take the most amount of time to mature: good documentation, clear usage, etc., just need a lot of back and forth between the user and the developer. You are a user, so please consider providing this back and forth. If you keep your complaints in the abstract level of "But at least the documentation and the way how error messages are designed could be improved.", then you won't see much improvement. Be specific and active, and guide the developers to the better version that you want. 

---

> In MATLAB, I can use vectors and matrices with the same functions. 

Sure, but in MATLAB you don't get the performance of Julia. `Dataset` exists because it makes computations faster. Such performance considerations will naturally bring into existence other things besides only vectors and matrices.

> But I would expect that “Dataset” should be something more standard.
> Would it be possible to promote the Dataset type for use other Julia packages to have it like a standard for similar use cases? 

This is the point where truly understanding multiple dispatch would dissolve this issue. Due to the system, it is natural that different groups would want their own types. I can explain this in further detail but it would take several paragraphs and I want to keep it short, feel free to open something up on discourse though and I'll answer.

Besides, your standarizing suggestion is unecessary for the moment. You can do `Matrix(dataset)` and `Dataset(matrix)`, or `x, y, z = columns(dataset)`. Instead of having several groups trying to use the same Type, which is counter-productive in the presence of multiple dispatch, we in Julia instead provide easy to use interfaces to transform between different types. It isn't as big of a deal as you claim it to be. After all `Matrix(dataset)` is less than a line of code.

>  In MATLAB this (almost) cannot happen because it is simple by design.

I couldn't disagree more with that statement. What I believe is that this doesn't happen for three different reasons: 1) MATLAB is simply not as powerful as Julia to allow multiple dispatch and the freedom that comes with it, 2) MATLAB is centralized, 3) MATLAB has matured to the point where some data structures have become the community norm, while Julia is not at that point yet. I don't think that "MATLAB is simpler" than Julia. Simpler would mean **same amount of capabilities**, but easier to achieve. MATLAB is nowhere near having the same capabilities though, the absence of Multiple Dispatch is evidence of that.

> If 90% of the users have difficulties in a first try to apply a function correctly, because the input should be of a certain format or data type, then perhaps it is better to re-design the code in a way to have it less error prone.

I absolutely agree. In many of my projects I actively ask the community what they believe the design of a next step should be, a bright example being Agents.jl. However there is one proviso: I'm only willing to consider people that have read the documentation. If *after* reading the documentation they still have difficulties in a first try to apply a function correctly, then clearly there are crucial improvements to be made. This is just my personal opinion however. Other developers might think that designing such a "magical" interface that doesn't need documentation is worth their time. So far I haven't seen such a person contributing to DynamicalSystems.jl.

> (perhaps it [DynamicalSystems.jl] could still be improved at some places, in particular in the function documentation that is available in the Juno documentation window).

Then please open an issue! In fact, open _many_ of them! For anything you can think it can be improved, you should open an issue! That's what drives open source projects by a large part.

> I would like to have a similar good documentation in general and also with many examples and how to use them in different situations.

Same answer, open issues! Make your concerns known!

pucicu:
I can understand this point and fine with it. But it might be good to have in the docstring a hint that it in some literature it has differing names. For example in the recurrence literature, the term “fixed amount of nearest neighbours” is quite common, etc.

heliosdrm:
Changed the types to  `WithinRange` and `NeighborNumber` as requested, but retained the alias `FAN` as convenience abbreviation as used in the literature. The relevant paragraph in the docstring now is like this:

> The parametrized constructor `RecurrenceMatrix{NeighborNumber}` creates the recurrence matrix
with a fixed number of neighbors for each point in the phase space, i.e. the number
of recurrences is the same for all columns of the recurrence matrix.
In such case, `ε` is taken as the target fixed local recurrence rate,
defined as a value between 0 and 1, and `scale` and `fixedrate` are ignored.
This is often referred to in the literature as the method of "Fixed Amount of Nearest Neighbors"
(or FAN for short); `RecurrenceMatrix{FAN}` can be used as a convenient alias
for `RecurrenceMatrix{NeighborNumber}`.

Datseris (approved):

pucicu:
Hi,

thanks! I had to read the description twice to understand that it is referring to the _local_ recurrence rate. During first reading I was misunderstanding that it would be the (globally) fixed recurrence rate. But his might be my fault.

Best
Norbert

PR #104, opened 2020-11-29
fix return type of transitivity

heliosdrm:
When the input was non-square, it returned `false`, whereas normally it should return a float. Alternatively, it could be changed to return `NaN`


Datseris:
it should return `NaN`, `0.0` is a valid value for transitivity

heliosdrm:
Now it returns `NaN`.

Issue #106, opened 2020-11-30
unexpected behaviour for arrays as input


pucicu:
I have imported data using `readdlm`, e.g., into variable _x_. The result is a `1001×5 Array{Float64,2}`. When I use, e.g., `x1 = x[:,5]` (resulting in `1001-element Array{Float64,1}`), then I can calculate `R = RecurrenceMatrix(x1, ε; fixedrate = true)`. But it does not work when I use the entire array _x_ , i.e., `R = RecurrenceMatrix(x, ε; fixedrate = true)`. I get this error:
`ERROR: MethodError: no method matching recurrence_matrix(::Array{Float64,2}, ::Euclidean, ::Float64, ::Val{true})`

This is completely unexpected.

Datseris:
Multidimensional data in the DynamicalSystems.jl ecosystem are expected in the form of `Dataset` and not `Matrix`. This is explained here: https://juliadynamics.github.io/DynamicalSystems.jl/dev/#Getting-started-1  as well as here https://juliadynamics.github.io/DynamicalSystems.jl/dev/embedding/dataset/ .

So, you have to give `Dataset(your_matrix)` to the function.

Hm, but I now checked the docstring of `RecurrenceMatrix` and it says that it can accept a `Matrix`. That's unfortunate, it is a remnant of a previous version. I (personally) don't agree with adding an automatic conversion from matrix to dataset in `RecurrenceMatrix`, because such conversions must be done manually in every single function of the ecosystem. I find more reasonable to use the expected form instead.

pucicu:
ok, I see. this increases of course the complexity of the user interface, but I understand that you have good reason for this.

therefore, I suggest to add a corresponding hint when this error is catched. instead of the auto-conversion from array to dataset.

Datseris:
> but I understand that you have good reason for this.

yes, the dimensionality of datapoints of `Matrix` is not known by type, which leads to inferior performance down the line.

heliosdrm:
Closed by #107: matrix inputs work now with all functions, but issue a deprecation warning; in the future it should be the other way round - throw an error with an informative message informing of the correct call.

PR #107, opened 2020-12-01
Deprecate AbstractMatrix inputs

heliosdrm:
To deal with #106 and the last comments on #102 

With this, using `julia --depwarn=yes`:

```julia
julia> using RecurrenceAnalysis
julia> x=randn(100, 2);
julia> RecurrenceMatrix(x,0.4)
┌ Warning: `RecurrenceMatrix(x::AbstractMatrix, ε; kwargs...)` is deprecated, use `RecurrenceMatrix(Dataset(x), ε; kwargs...)` instead.
```

But the calculations go as if the matrix had been converted to a `Dataset`. (By default, in Julia 1.5+ the deprecation warning is hidden).

If this is approved, `src/deprecate.jl` could be extended later on for `RecurrenceMatrix{FAN}` etc.


heliosdrm:
My doubt is: do we want to keep the deprecation warning as it is by default (silent unless you activate the option `depwarn=yes`, which most people won't do), or is it better to make it more prominent, and always throw the warning to instruct users that `Dataset` must be used for multi-dimensional data?

Datseris:
I think the depwarning always has to be thrown. How did you make it "not being thrown by default"? When I use `@deprecate` or `@warn` the are always shown when starting Julia the default way.

src/matrices.jl
```diff
@@ -68,7 +68,7 @@ Objects of type `<:AbstractRecurrenceMatrix` are displayed as a [`recurrenceplot
 The recurrence matrix is a numeric representation of a "recurrence plot" [^1, ^2],
 in the form of a sparse square matrix of Boolean values.
 
-`x` must be a `Vector` or a `Dataset` or a `Matrix` with data points in rows
+`x` must be a `Vector` or a `Dataset`
 (possibly representing and embedded phase space; see [`embed`](@ref)).
```
Datseris, line 72: spotted a typo here: "possibly representing **an** embedded space"

heliosdrm:
> How did you make it "not being thrown by default"? When I use `@deprecate` or `@warn` the are always shown when starting Julia the default way.

I did not make it. It's the default as of Julia 1.5:
https://docs.julialang.org/en/v1/base/base/#Base.@deprecate

Datseris:
Okay then it is an easy change. Instead of using `@deprecate` we can instead use `@warn` inside the function call.

heliosdrm:
`@warn` would repeat the message every time matrices are used as inputs, whereas `@deprecate` would show the message only the first time it happens. Also, according to [the answers in Discourse](https://discourse.julialang.org/t/why-is-depwarn-no-by-default/51094/11) in 1.6 we would be able to force the deprecation message.

Should we make it a regular warning, instead of a deprecation warning, still?

Datseris:
> `@warn` would repeat the message every time matrices are used as inputs

That's what I always do actually. Don't know if it is good practice or not, but if I know there is some code that will break in the next release, I'd rather press the user to change it as soon as possible instead of passively letting it go after 1 warning.

PR #108, opened 2020-12-09
changes to `rqa` with deprecations

### LINKED ISSUE #105: Redefine the `rqa` function
The `rqa` function has two issues right now:

* With the addition of `TRANS`, when the input is a nonsquare matrix it return `TRANS=false` with a warning. Maybe it would make more sense to skip that parameter.
* It is type-unstable: it returns a `NamedTuple` with a different set of elements depending on the keyword argument `onlydiagonal`.

(Breaking) proposal: change it to two different functions depending on whether diagonal-related parameters are wanted or not. `TRANS` could be skipped for `CrossRecurrenceMatrix` inputs, which are not guaranteed to be square.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/94993369-redefine-the-rqa-function?utm_campaign=plugin&utm_content=tracker%2F28405688&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F28405688&utm_medium=issues&utm_source=github).
</bountysource-plugin>

heliosdrm:
Closes #105.

In order to facilitate the transition from `NamedTuple` to `Dict`, I have defined different methods of `rqa` (more details in the docstring):

* `rqa(NamedTuple, R...)` to obtain the output of the older version (as in 1.3).
* `rqa(Dict, R...)` to obtain the output of the planned future version.
* `rqa(RQA, R...)` to obtain the default output (same as `rqa(R...)`)

The `RQA` type just wraps a dictionary "inheriting" the following functions: `getindex`, `get`, `length`, `keys`, `values`, `collect`, `iterate`. More functions might be needed.

The rewriting of the `@windowed` macro only considers the planned API, i.e. `@windowed rqa ...` returns a dictionary of vectors, with no deprecation warnings or error messages if one tries to call the result as if it were a named tuple. Implementing that would be more complicated than in the "standard" `rqa` function. But I don't think that we should be compelled to do it, because although the macro *can* be used with the `rqa` function (and it actually works), that possibility was never documented - the documentation tells about the individual RQA functions, and about `RecurrenceMatrix` etc..

### DIFF
--- a/src/rqa.jl
+++ b/src/rqa.jl
@@ -92,25 +92,25 @@ refers to the fact that for two other vertices `j, k ∈ V` with
 `A_ij = A_ik = 1, A_jk = 1` also holds. In a general network, this is typically
 not the case for all vertices. Consequently, characterising the degree of
 transitivity (or, alternatively, the relative frequency of closed 3-loops, which
-are commonly referred to as triangles) with respect to some individual vertex or
+are commonly referred to as triangles) with respect to 
 the whole network provides important information on the structural graph
 properties, which may be related to important general features of the underlying
 system.
 
-The network transitivity averages the local transitivity or clustering
-coefficient
+The network transitivity measures the fraction of closed triangles with respect to
+the number of linked triples of vertices in the whole network:
+
 ```math
-\\mathcal{C} = \\frac{\textrm{number of triangles including vertex i}}{\textrm{number of triples centred on vertex i}}
+\\mathcal{T} = \\frac{3 \\times \\textrm{number of triangles in the network}}{\\textrm{number of linked triples of vertices}}
 ```
-over the all nodes in the network [^Boccaletti2006]:
+or given the adjacency matrix `A`
 ```math
-\\mathcal{T} = \\frac{trace(R^3)}{\\sum R^2}
+\\mathcal{T} = \\frac{trace(A^3)}{\\sum(A^2) - trace(A^2)}
 ```
 
 ## References
 
 [^Donner2011]: R.V. Donner *et al.*, [The geometry of chaotic dynamics — a complex network perspective, Eur. Phys. J. B 84, 653–672 (2011)](https://doi.org/10.1140/epjb/e2011-10899-1)
-[^Boccaletti2006]: S.Boccaletti *et al.*, [Complex networks: Structure and dynamics, Physics Reports Volume 424, Issues 4–5 (2006)](https://doi.org/10.1016/j.physrep.2005.10.009)
 """
 function transitivity(R::ARM)
     if size(R, 1) ≠ size(R, 2)
@@ -118,8 +118,14 @@ function transitivity(R::ARM)
         return NaN
     end
     R² = R.data * R.data
-    R³ = R² * R.data
-    trans = LinearAlgebra.tr(R³) / sum(R²)
+    numerator = zero(eltype(R²))
+    for col = 1:size(R,2)
+        rows = view(rowvals(R), nzrange(R,col))
+        for r = rows
+            numerator += R²[r, col]
+        end
+    end
+    trans = numerator / (sum(R²) - LinearAlgebra.tr(R²))
 end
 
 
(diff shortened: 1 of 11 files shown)


heliosdrm:
I suggest the following roadmap:

1. Release a version with the different methods as in this PR: (with `NamedTuple` for backwards compatibility, `RQA` to help in the transition, and `Dict` for the future). Users can choose any of the three, and default is with `RQA`.
2. Drop the method with `NamedTuple`, and make the `Dict` method the default, moving the `RQA` method to the list of deprecated.
3. Drop the method with `RQA`, and leave the simple `Dict` output as the only option.

(Steps 2 and 3 might change if in the near future we discover some reason to stick with a custom `RQA` type, that might have more features than a simple dictionary. In that case, we would only remove its -already deprecated- `getproperty` method.)

src/rqa.jl
```diff
@@ -410,29 +434,33 @@ Using this function is much more efficient than calling all individual functions
...
+All the parameters returned by `rqa` are `Float64` numbers,
+even for parameters like `:Lmax`, `:Vmax` or `:NMPRT` which are integer values.
 In the case of empty histograms (e.g. no existing vertical lines
 less than the keyword `lminvert`) the average and maximum values
-(`L`, `Lmax`, `TT`, `Vmax`, `MRT`)
-are returned as `0.0` but their respective entropies (`ENTR`, `VENTR`, `RTE`)
+(`:L`, `:Lmax`, `:TT`, `:Vmax`, `:MRT`)
+are returned as `0.0` but their respective entropies (`:ENTR`, `:VENTR`, `:RTE`)
 are returned as `NaN`.
+`:TRANS` is only returned when the input is a square matrix.
```
Datseris, line 463: Now that I read this, isn't it more consistent to return `NaN` for TRANS of non-square, similarly with how the entropies are NaN when no other sensible value is in order?

Datseris (approved):

Datseris:
Sounds good to me!

src/rqa.jl
heliosdrm: The situation of TRANS is not exactly the same as that of the entropies. Entropies return `NaN` when there are no lines (longer than the minum length) in the matrix, which is a "legitimate" case. (The formula of the Shannon entropy (`-sum(x .* log.(x))`) would give zero for an empty histogram, but we coerce `NaN` in order to distinguish that case from that where all the lines have the same length, which also returns zero.) On the other hand, there is no point in trying to calculate TRANS for non-square matrices (its definition is related to a single set of vertices, so the matrix must have the same number of columns and rows).

Actually, I think that the `transitivity` function could (should?) throw an error (with the same informative message) instead of a warning, when it is called with non-square matrices.

Datseris: if it throws an error what happens with calling `rqa` for non-square matrices?

heliosdrm: Nothing happens: since dictionaries are mutable, the dictionary is initially created without the `:TRANS` key, and it is added only if the matrix is square.

Datseris: a okay sure.

Datseris:
ops latest merged failed some tests.

heliosdrm:
Take care when a version number is given to this. `rqa` as coded here does not return a tuple as in previous versions (you must use the deprecated `rqa(NamedTuple, R, ...)` to obtain that result). So this might be considered a breaking change!

Datseris:
But the access is the same right? You can still do `returned_result.DET`? That's the reason I wouldn't increment major version with this change.

heliosdrm:
Yes, you can do `returned_result.DET`, which is I think the most likely error to be found in code that expects a named tuple. But there are other differences in the interface between named tuples and dictionaries that are not worked around by the `RQA` type, namely:
* Named tuples accept numeric indices, e.g. `returned_result[1]`, whereas `RQA` can only be indexed as its wrapped dictionary.
* If you iterate over an `RQA` you will get key-value pairs, as in dictionaries - with named tuples you would only obtain the values.

The order of elements in the named tuple returned by previous versions of `rqa`is rather arbitrary, so I don't expect that many people should stumble on those differences, but there they are, anyway.

(Although on second thought, if someone's code has problems with those differences, it would have been affected by changes in previous minor releases too, as when TRANS was added as *second* item of the result, moving the position of the rest of parameters one index forward in the list. So after all it is not more breaking than previous minor releases were.)

PR #110, opened 2020-12-13
fix transitivity

### LINKED ISSUE #109: Unexpected results of `transitivity`
Where does this formula for transitivity come from?

https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/blob/2511090f56104f20416667124c4346407462057b/src/rqa.jl#L106-L108

I didn't find it in the cited papers, an it does not produce the same results as `global_clustering_coefficient` from LightGraphs.jl, although the concept cited in the docstring does coincide. (In case of doubt, I have tried with hand-made tiny examples, and the result of LightGraphs looks correct, but the output of `transitivity` here does not.)

To reproduce the inconsistency, try `SimpleGraph(R.data) |> global_clustering_coefficient` and compare with `transitivity(R)`.

(By the way, transitivity would normally be calculated with adjacency matrices that do not have elements in the diagonal - I am addressing that in the branch `graphs` - WIP)

heliosdrm:
Fix #109

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RecurrenceAnalysis"
 uuid = "639c3291-70d9-5ea2-8c5b-839eba1ee399"
 repo = "https://github.com/JuliaDynamics/RecurrenceAnalysis.jl.git"
-version = "1.3.1"
+version = "1.3.2"
 
 [deps]
 DelayEmbeddings = "5732040d-69e3-5649-938a-b6b4f237613f"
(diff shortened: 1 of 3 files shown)


heliosdrm:
Also modified the way of computing the numerator: we only need the diagonal elements of `R^3`, which can be obtained more efficiently by summing the elements of `R^2` which are also in `R`. The proposed implementation saves about 30% and 50% of time and memory.

Datseris:
Shouldn't we add a test for this based on your comments from #109 ?

pucicu:
very good idea! I support this!

Datseris:
Don't forget to up Project.toml: PRs like this should be released after merge.

Issue #23, opened 2018-12-31
Functions for analysis of (cross-)recurrence networks


heliosdrm:
This includes the calculation of:

- [ ] Adjacency matrix (based on `recurrencematrix`)
- [ ] Edge density (based on `recurrencerate`)
- [ ] Global clustering coefficient or network transitivity
- [ ] Average path length
- [ ] Global efficiency

pucicu:
Hi,

would be good to have also the Laplace matrix and the Eigenvalue spectrum.
See:

D. Eroglu, N. Marwan, S. Prasad, J. Kurths: Finding recurrence networks' threshold adaptively for a specific time series, Nonlinear Processes in Geophysics, 21, 1085–1092 (2014). DOI:10.5194/npg-21-1085-2014 <http://dx.doi.org/10.5194/npg-21-1085-2014>

Best wishes for the new year!
Norbert

heliosdrm:
Going back to old issues, and I found this one, half-forgotten. Now we already have the `transitivity` function and the TRANS parameter from the `rqa` function, although without having the type `AdjacencyMatrix <: AbstractRecurrenceMatrix` that was suggested originally in this issue.

In #105 and #108 we were discussing the problem of TRANS not being applicable to non-square matrices. If we recover the idea of creating a specific type for `AdjacencyMatrix`, perhaps we could make `rqa` return a different set of parameters for that type of matrices, including transitivity and the rest of parameters mentioned above (edge density, average path length, global efficiency).

Datseris:
Yes, I agree with this suggestion. Define the type, however I would call it `AdjacencyRecurrenceMatrix`, as adjacency by itself might mislead in the direction of e.g. LightGraphs.jl. 

Now that I mention this, have you checked LightGraphs? They might already have functions for all the quantities you mention.

heliosdrm:
[...]

PR #112, opened 2021-01-02
Support for recurrence networks based on LightGraphs

heliosdrm:
This includes:
* `SimpleGraph` constructor for `AbstractRecurrenceMatrix` (which only works for symmetric matrices).
* `transitivity`, `averageclustering` and `averagepath` functions (dispatching on `SimpleGraph`)
* Reexported from LigthGraphs: `SimpleGraph`, `degree`, `density`, `local_clustering`, `closeness_centrality`, `betweenness_centrality`, `laplacian_matrix` and `laplacian_spectrum`.

I'm not sure about the reexported functions: the selection is based on what has been mentioned in #23, and the list of parameters in Chapter 4 of Webber and Marwan's RQA book, but I don't know if that's a sensible selection, or looks too arbitrary. Taking into account the confusion with the concept of "global clustering coefficient", I have deliberately left out the function `global_clustering_coefficient` from LighGraphs, but defined `transitivity` as an alias for it, and also made `averageclustering` to account for the alternative calculation.

Regarding what was already in the package, and following what [I suggested in #109](https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/issues/109#issuecomment-744044979):
* `transitivity(::AbstractRecurrenceMatrix)` is moved to `deprecated.jl`.
* `rqa` does not supply the TRANS parameter anymore.

When I wrote that comment I was thinking in adding an option to keep the main diagonal (LOI) in the creation of the `SimpleGraph` from recurrence matrices, as if there were self-connected nodes in the network. But I have removed that option, because I think it would be a bit odd, and without it the constructor is closer to the one given in LightGraphs for `SimpleGraph(::AbstractMatrix)`.

Remaining doubts:

* The deprecated version of `transitivity(x::AbstractRecurrenceMatrix)` is left unchanged; this means that the result of the recommended call, `transitivity(SimpleGraph(x))` is different, because normally `x` will contain the main diagonal (LOI) but the adjacency matrix of the graph does not. Should the deprecated function just call the `transitivity(SimpleGraph(x))` to make the results consistent (within this version, but not with the previous ones)?
* Should TRANS be kept as part of the results of `rqa`? I have removed it because it was rather an outlier: a parameter coming from a different way of analyzing recurrences, whose calculation does not share anything with the rest of parameters (so there is no real gain in performance by calculating it together with the others), which can only be calculated if the matrix is square... I like it better outside `rqa`, but such a change might be annoying for some current users...


Datseris:
> I'm not sure about the reexported functions

Me neither. Many exported names are based on networks, and all these names will also invade the exports of DynamicalSystems.jl. What I'd rather we do is that we have an `rna` function, returning a dictionary with keys the names of these functions.

> Should TRANS be kept as part of the results of rqa?

Nope, this is totally fine.

heliosdrm:
> What I'd rather we do is that we have an `rna` function, returning a dictionary with keys the names of these functions.

Yes, that would be similar to what is done for conventional RQA. Although except for that, there is not a real advantage of providing an "all-in-one" function for analysis of networks. In the case of RQA, there is a well-established, relatively small set of parameters that are usually calculated, and the heavier part (defining the histogram of diagonal or vertical structures) is shared by many of those parameters, so the `rqa` function is a tool to save time.

In the case of networks, we are simply calling functions form LightGraphs.jl, so there is little gain in having a function that calculates various parameters. Unless you want *exactly* the same set of parameters provided by such a function, you'd better call the individual functions you need, and organize the results the way you prefer (in a dictionary, in different variables...).

I'm not that acquainted with the analysis of networks. So, unless Norbert or someone else suggests some "standard" set of parameters that might be worth to provide together, I think that the best option is to include details in the documentation of what functions from LightGraphs can be called to obtain parameters that are usually found in the literature.

In such case, I wonder if the alias of `transitivity` for `LightGraphs.global_clustering_coefficient` still makes sense, or we should just explain plainly in the documentation the nomenclature issue that was discussed in #109.

heliosdrm:
Also, instead of exporting individual functions from LightGraphs.jl that might conflict with other names exported by DynamicalSystems.jl (`degree`, `density`, etc.), we could export the module `LightGraphs` itself. That way, users would only need to type `LightGraphs.degree` etc.

In many cases, the second option will be more convenient. But not if the user is also `using LightGraphs` in the same environment for any reason. In that case,

Datseris:
I don't think we should re-export stuff from lightGraphs.jl. I still stand that an `rna` function is much better than exporting many names, some of which are direct duplicate exports from LightGraphs. The `rna` keeps things clean. On the other hand, it does compute everything every time which might not be optimal.

Perhaps the best way to proceed here is to simply create a new documentation page that calculates all quantities in `rna` once (i.e. exlicitly shows how). So a new page that uses both RA and `LightGraphs` and simply calculates all quantities for a given matrix and presents them. As a result, I think we shouldn't export _anything_ extra, not even `SimpleGraph`, as it comes into scope anyways when using LightGraphs.

A new documenation page anyway has to be written where DynamicalSystems.jl docs are hosted, otherwise I doubt anyone besides us 4 will ever discover these new things.

heliosdrm:
So, let me put it in other words to see if I understood the proposal right: we provide the function `rna` that takes a recurrence matrix as input, and does all the stuff with LightGraphs behind the scenes (nothing re-exported from that package); but we also give a detailed documentation page that tells how one can achieve those results by `using LightGraphs` and calling the corresponding functions - e.g. to be more efficient if not all the results of `rna` are wanted, or to calculate other parameters available with the many other functions that LightGraphs has.

Then, the `SimpleGraph(::AbstractRecurrenceMatrix)` constructor will still be defined, although it will only be accessible directly if you have brought `LightGraphs` into scope. That won't be needed for users who are happy with the set of results given by `rna`. Right?

I'll browse some papers and packages for other languages (e.g. pyunicorn) to make a proposal of the results that might be provided by `rna` (LightGraphs gives so many possibilities...). But if Norbert's advice here will also be appreciated. :slightly_smiling_face:

pucicu:
Hi,

Have a nice new year!

I’m not sure if I understood it right.

But it would be good to have it simple. As a user, I would expect that I can derive the network measures in a similar way as the RQA measures. Do you have an example how it would be designed now? In Python it is something like

`rp = RecurrencePlot`
`rp.laminarity()`
`rp.transitivity()`

If there is no other way, it could also be to first get the RNA and have it

`rna.transitivity()`

And yes, it should be called _transitivity_ and not not _global clustering coefficient_ because the latter is the average of the local clustering coefficient.

Other network measures could be average path length, assortativity, and closeness.

Thank you very much!

Best wishes
Norbert

Datseris:
@heliosdrm exactly, you got it down perfect.

@pucicu here you would do:
```julia
x = rna(R)
x["transitivity"]
```
since `x` will be a dictionary. to directly get only a single measure instead you would do
```julia
g = SimpleGraph(R)
z = network_measure_function_from_LightGraphs(g)
```

pucicu:
ok, this looks good. thanks!

heliosdrm:
Thanks! I have it clear now.

Also:

>  to directly get only a single measure instead you would do
> 
> ```julia
> g = SimpleGraph(R)
> z = network_measure_function_from_LightGraphs(g)
> ```

Just add that to do it this way, you'll need having `LightGraphs` added (it'll be implicitly installed with RecurrenceAnalysis, but not visible by default), and type `using LightGraphs` before those lines. That will be duly explained in the documentation.

heliosdrm:
The function `rna` gives the following global parameters: edge density, transitivity, average shortest path, and diameter. Assortativity will be added as soon as that function comes [available in LightGraphs.jl](https://github.com/JuliaGraphs/LightGraphs.jl/issues/1508)

heliosdrm:
ready to be revised

Datseris (approved):
Great, very clean

PR #117, opened 2021-05-21
Map Inf path lengths to zero

### LINKED ISSUE #116: RNA `:averagepath` returns Inf with disconnected nodes
RNA `:averagepath` should assign a value of zero for each disconnected node as discussed in [Donner, 2010](https://iopscience.iop.org/article/10.1088/1367-2630/12/3/033025). Because the path length has been defined as the 1 / closeness centrality [cc], when cc == 0, the path length becomes Inf and so does the mean.
Will post a PR soon.

gerrygralton:
Issue opened in #116.

Closes #116

### DIFF
--- a/src/rna.jl
+++ b/src/rna.jl
@@ -43,7 +43,8 @@ function rna(args...; kwargs...)
     return Dict{Symbol, Float64}(
         :density => density(graph),
         :transitivity => global_clustering_coefficient(graph),
-        :averagepath => mean(map(x-> isinf(x) ? 0. : x, 1 ./ closeness_centrality(graph))),
+        :averagepath => sum(map(x-> x == typemax(x) ? 0. : x,
+                        floyd_warshall_shortest_paths(graph).dists)) / (nv(graph) * (nv(graph) - 1)),
         :diameter => diameter(graph)
     )
 end
--- a/src/rna.jl
+++ b/src/rna.jl
@@ -43,7 +43,7 @@ function rna(args...; kwargs...)
     return Dict{Symbol, Float64}(
         :density => density(graph),
         :transitivity => global_clustering_coefficient(graph),
-        :averagepath => mean(1 ./ closeness_centrality(graph)),
+        :averagepath => mean(map(x-> isinf(x) ? 0. : x, 1 ./ closeness_centrality(graph))),
         :diameter => diameter(graph)
     )
 end
--- a/src/rna.jl
+++ b/src/rna.jl
@@ -43,8 +43,22 @@ function rna(args...; kwargs...)
     return Dict{Symbol, Float64}(
         :density => density(graph),
         :transitivity => global_clustering_coefficient(graph),
-        :averagepath => sum(map(x-> x == typemax(x) ? 0. : x,
-                        floyd_warshall_shortest_paths(graph).dists)) / (nv(graph) * (nv(graph) - 1)),
+        :averagepath => averagepath(graph),
         :diameter => diameter(graph)
     )
+end
+
+"""
+    averagepath(graph)
+
+Calculates average minimum path length for a SimpleGraph.
+Uses Donner, 2010, Eq. 26 on pg. 18.
+"""
+function averagepath(graph::SimpleGraph)
+    num_verts = nv(graph)
+
+    dist_mat = floyd_warshall_shortest_paths(graph).dists
+    @. dist_mat[dist_mat == typemax(dist_mat)] = 0.
+
+    return sum(dist_mat) / (num_verts * (num_verts - 1))
 end
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # RecurrenceAnalysis.jl News
 
+## v1.5.2
+- New, faster formula for `rna[:averagepath]` allowing disconnected graphs.
+
 ## v1.5.1
 - Minor documentation fixes.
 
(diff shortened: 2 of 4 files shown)


Datseris:
Thank you :) @gerrygralton Could you please provide a unit test for this scenario? You should put the test in one of the files in the `test` folder. Reading any file will guide you on how test work.

@heliosdrm  or  @hkraemer would you kindly review and approve the PR? I do not have the mathematical background :)

heliosdrm:
Thanks! If you have a specific example of a network with disconected nodes that you want to test, either defined as a `LightGraph` or by its adjacency matrix, you may add it in the last block of `test/smallmatrix.jl`, where there are other examples that you can follow.

gerrygralton:
Hi, 
Thanks for having a look at this! In 127dad8a8de6679d8e09f352e7dd5dab2e1aec15 I've changed the formula slightly to remove the LightGraphs.closeness_centrality dependency because that formulation really doesn't like disconnected graphs.
It also has the advantage of being faster. Using the 10x10 example from test/small_matrix.jl:

New:
    
    BenchmarkTools.Trial: 
      memory estimate:  2.67 KiB
      allocs estimate:  5
      --------------
      minimum time:     929.786 ns (0.00% GC)
      median time:      957.500 ns (0.00% GC)
      mean time:        1.104 μs (5.17% GC)
      maximum time:     101.671 μs (98.23% GC)
      --------------
      samples:          10000
      evals/sample:     28
  
Old:

    BenchmarkTools.Trial: 
      memory estimate:  20.08 KiB
      allocs estimate:  165
      --------------
      minimum time:     11.829 μs (0.00% GC)
      median time:      12.584 μs (0.00% GC)
      mean time:        15.399 μs (8.92% GC)
      maximum time:     3.706 ms (98.96% GC)
      --------------
      samples:          10000
      evals/sample:     1

I have also added some tests in 989bd14dd8029873409c986c24f81662a350a91b.

src/rna.jl
```diff
@@ -43,7 +43,8 @@ function rna(args...; kwargs...)
     return Dict{Symbol, Float64}(
         :density => density(graph),
         :transitivity => global_clustering_coefficient(graph),
-        :averagepath => mean(map(x-> isinf(x) ? 0. : x, 1 ./ closeness_centrality(graph))),
+        :averagepath => sum(map(x-> x == typemax(x) ? 0. : x,
+                        floyd_warshall_shortest_paths(graph).dists)) / (nv(graph) * (nv(graph) - 1)),
```
Datseris, line 47: I do not think this is clear source code. So many operations in one line of code, some of which are not transparent to me. Can you please clarify the source a bit, and separate this into more steps?

gerrygralton: Sure, the source for this is in [Donner, 2010](https://iopscience.iop.org/article/10.1088/1367-2630/12/3/033025) pg. 18 eq. 26 (ref 1 in the rna.jl file).

I agree, that's a dense function. All it does is get the path lengths from LightGraphs, maps the Inf values to zero (as in Donner 2010) and then divided the sum by (N* (N-1)). If I break it down into multiple lines where would you like function to go?

Datseris: Just define a function after `rna` and call this function in the `:averagepath` mapping. Especially since you have a specific equation `Donner, 2010 pg. 18 eq. 26`, this should be referenced as a comment in the function :)

gerrygralton: OK. I've implemented that. New benchmarking is only slightly slower.

    BenchmarkTools.Trial: 
      memory estimate:  2.06 KiB
      allocs estimate:  8
      --------------
      minimum time:     1.060 μs (0.00% GC)
      median time:      1.125 μs (0.00% GC)
      mean time:        1.348 μs (5.33% GC)
      maximum time:     207.975 μs (98.31% GC)
      --------------
      samples:          10000
      evals/sample:     10

src/rna.jl
```diff
@@ -43,7 +43,22 @@ function rna(args...; kwargs...)
...
+Uses Donner, 2010, Eq. 26 on pg. 18.
+"""
+function averagepath(graph::SimpleGraph)
+    num_verts = nv(graph)
+
+    dist_mat = floyd_warshall_shortest_paths(graph).dists
+    @. dist_mat[dist_mat == typemax(dist_mat)] = 0.
+
+    return sum(dist_mat) / (num_verts * (num_verts - 1))
+end
```
Datseris, line 64: Should we consider to add this function to LightGraphs.jl directly...?

gerrygralton: Probably, someone had a shot here: https://github.com/JuliaGraphs/LightGraphs.jl/issues/1259.

Datseris: On second thought, we leave it here. :)

Datseris:
Thanks, can you please adjust Project.toml version and CHANGELOG.md accordingly?

gerrygralton:
> Thanks, can you please adjust Project.toml version and CHANGELOG.md accordingly?

Uh, there is no CHANGELOG.md...

Datseris:
Woooooooooooops sorry about that, let's start one then. Just provide the note that changelog starts being kept from current version.

gerrygralton:
Is NEWS.md the same thing?

Datseris:
Ah, sorry for the confusion @gerrygralton . Yeap, it is NEWS.md. I got confused, in some other packages it is called CHANGELOG.md...

gerrygralton:
@Datseris This new version hasn't been tagged because TagBot has turned itself off.

Datseris:
Oh, okay, do you know how to turn it on again?

PR #122, opened 2021-10-31
Allow `rqa` and associated functions to allow `SparseMatrixCSC`.

### LINKED ISSUE #92: `rqa` functions should also take in SparseArrays
This makes it easier to calculate metrics on segments of a recurrence matrix.

### LINKED ISSUE #121: Question on best place to allow modification to condition for recurrence.
I am working on creating a symbolic recurrence plot to add color to the recurrences. Essentially I just want it so a 1 only occurs when `x[i]=y[j]= some value`. I am just not sure if it makes more sense to just rewrite `recurrence_matrix` with this command and add an extra keyword to allow for a custom metric to be applied or if I need to create a brand new `RecurrenceMatrix` type. Or do I just need to modify get metric?

I currently have a very rough work around, but a majority of what I have coded is all the same that is already in this package besides a few lines, but I have not gotten to the analysis yet and a lot of the low level functions in rqa are still applicable. 

Solving #92 I think would also solve my problem though too.

rs7q5:
This PR should resolve issues #121 and #92 . The only potential thing would be if the default `theiler` window for `SparseMatrixCSC` should be 1 or 0. It is currently set to zero. If it should be set to 1 I think the only changes would be in `rr_denominator` and `deftheiler`.

Closes #92. Closes #121.

### DIFF
--- a/src/rqa/histograms.jl
+++ b/src/rqa/histograms.jl
@@ -100,9 +100,9 @@ function _linehistograms(rows::T, cols::T, lmin::Integer, theiler::Integer,
 end
 
 deftheiler(x::Union{RecurrenceMatrix,JointRecurrenceMatrix}) = 1
-deftheiler(x::CrossRecurrenceMatrix) = 0
+deftheiler(x::Union{CrossRecurrenceMatrix,SparseMatrixCSC}) = 0
 
-function diagonalhistogram(x::ARM; lmin::Integer=2, theiler::Integer=deftheiler(x), kwargs...)
+function diagonalhistogram(x::Union{ARM,SparseMatrixCSC}; lmin::Integer=2, theiler::Integer=deftheiler(x), kwargs...)
     (theiler < 0) && throw(ErrorException(
         "Theiler window length must be greater than or equal to 0"))
     (lmin < 1) && throw(ErrorException("lmin must be 1 or greater"))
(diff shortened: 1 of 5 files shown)


test/dynamicalsystems.jl
```diff
@@ -124,4 +124,16 @@ dict_keys = ["Sine wave","White noise","Hénon (chaotic)","Hénon (periodic)"]
...
+    rqapar_sparse = rqa(rmat.data, theiler=1, lmin=3, border=20)
+    rqadiag_sparse = rqa(rmat.data, theiler=1, lmin=3, border=20, onlydiagonal=true)
+    for p in keys(rqadiag_sparse)
+        @test rqapar_sparse[p]==rqadiag_sparse[p]
+    end
+
+    #check for erroring here with sparse matrices
+    @windowed recurrencerate(rmat.data,theiler=1) width=50 step=40
+    @windowed rqa(rmat.data,theiler=1) width=50 step =40
+    coordinates(rmat.data)
```
Datseris, line 138: there is no `@test` done here.

src/rqa/rqa.jl
```diff
@@ -349,8 +347,7 @@ of recurrence times [1].
 [DOI:10.1103/physreve.75.036222](https://doi.org/10.1103/physreve.75.036222)
 
 """
-nmprt(R::ARM; kwargs) = maximum(verticalhistograms(R; kwargs...)[2])
-
+nmprt(R::Union{ARM,SparseMatrixCSC}, kwargs...) = maximum(verticalhistograms(R.data;theiler=deftheiler(R), kwargs...)[2])
```
Datseris, line 350: I don't understand. How can this definition work? A `SparseMatrixCSC` does not have a field `.data`.

src/rqa/histograms.jl
```diff
@@ -100,9 +100,9 @@ function _linehistograms(rows::T, cols::T, lmin::Integer, theiler::Integer,
 end
 
 deftheiler(x::Union{RecurrenceMatrix,JointRecurrenceMatrix}) = 1
-deftheiler(x::CrossRecurrenceMatrix) = 0
+deftheiler(x::Union{CrossRecurrenceMatrix,SparseMatrixCSC}) = 0
```
Datseris, line 103: Not a good idea to group the default of a `CrossRecurrenceMatrix` with that of a `SparseMatrix`. Please add a third definition. Also, givcen that we do not know the background of the sparse matrix we should use the safer option and give it a default theiler of 1.

test/dynamicalsystems.jl
rs7q5: I was simply checking if it were to error out similar to  what is now line 113 in this file. I have added a test for coordinates though. Considering `rmat` has not changed, I figured you mostly just needed to check it did not error out and I did test that the values coming out would be the same since I have already added tests that the `rqa` functions work with sparse arrays, so just need to check that the `@windowed` macro worked.

I went ahead and added two `@test` though for `coordinates` and the `:RR` computed in 136 and 137.

src/rqa/rqa.jl
rs7q5: That should only be `R`, but your point does still stand. I'm not sure. But I did change it to simply just `R` so the right type does get fed through. 

Must be something internally with SparseMatrixCSC falling back on something? I have no idea.

Issue #125, opened 2021-12-05
Display or return rqa results in the order shown in the docs


rs7q5:
I would like the rqa results to be shown or returned in the order that they are inserted because as it sits it currently makes sense (RR, then diagonal measures then vertical measures if desired) and it would also be consistent with how it is presented in the docs. 

I understand why the exported version of it is a Dict, but to achieve the desired result the quickest way I can think of is to allow the rqa type specified to be an AbstractDict to allow for an OrderedDict to be specified and without adding a dependency for different types of collections.

In other words the proposed change is to change the rqa function to 

`function rqa(::Type{DT},R;kwargs...) where {DT<:AbstractDict} end`

I would also be content with just editing Base.show, but I wasn't sure which would be preferred.

Datseris:
What happens if you just do `OrderedDict(rqa(...))`?

Datseris:
but sure, you can just do a pr that uses AbstractDict.

rs7q5:
If you just do `OrderedDict` you will get the default order that the dictionary displays it as because `OrderedDict` is the insertion order. 

I'll submit a PR request though. thanks!

PR #127, opened 2021-12-05
`rqa` can return an `AbstractDict` object.

rs7q5:
Solves #125.


Datseris:
can you also add this info to the documentation string?

rs7q5:
Sorry for all of the commits. I thought I did something weird with the commands via `git`. The change should be there now though.

Issue #98, opened 2020-10-12
Skeletonization of RPs and border effect correction.


hkraemer:
Hi there,

the finite size of a recurrence plot (RP) can cause border effects in the RQA-quantifiers. Also the samplingrate of the data and the chosen recurrence threshold selection method (`fixed`, `fixedrate`, `fan`) plays a crucial role. They can cause the thickening of diagonal lines in the RP. Both problems lead to biased line-based RQA-quantifiers and is discussed in [Kraemer & Marwan, Phys. Lett. A (2019)](https://publications.pik-potsdam.de/pubman/faces/ViewItemOverviewPage.jsp?itemId=item_23376). The proposed methods are implemented in Matlab and are [open source](https://zenodo.org/record/3384892#.X4RgL1ngrKI). 

According to the above mentioned two major problems, there are two methods needed, which will fix them. In the first case one simply needs to supress lines in the diagonal line length histogram, which cross the borders of the RP in an appropriate way (in the paper we compared four different ways, but the `kelo=keep longest line`-approach will do - see attachement). The second problem is way trickier to tackle. The problem corresponds to applying a Theiler window, but not only for the fiducial point on the trajectory, but also for its nearest neighbours falling within the recurrence threshold. We proposed a skeletonization scheme, which seems to work well (`rp_diagonal`, attached). 

Basically, we 

- first transform the RP by rotating it by 90°. This way the diagonal lines become horizontal lines. That might also speed up computation.

-  We pick the longest line from the diagonal line length histogram and iterate over each point of this line. At each point, we check if there is a neighbouring point/line. If yes, we repeat the same operation on this line until we do not find any neighbouring points and recursively delete all these adjacent lines from the histogram and the RP.

- Then we pick the second longest line from the histogram and repeat the mentioned procedure.

- We do this until there is no more line to pick from the histogram.

This leaves us with a RP only consisting of diagonal lines of thickness 1. Since the algorithm is not so complicated and basically consists of iterating over a histogram over and over agai,n I assume it could be way more performant than the Matlab implementation, what do you think?

I would be very grateful to get any help, while implementing this in Julia.

Cheers,
Hauke

hkraemer:
[...]

PR #128, opened 2021-12-14
Skeletonization

hkraemer:
Hi,
this is answering #98 in parts. Here I implemented skeletonization of RPs, which tackles the bias in RQA due to tangential motion. It is based on [this paper](https://www.sciencedirect.com/science/article/abs/pii/S0375960119308448) and already discussed in #98 . The tests have been adjusted and the new functionality is compressed into `/matrices/skeletonization.jl`

I tried a lot of stuff for optimizing it, but now I do not know how to increase performance anymore. The basic steps of the algorithm are:

- transform the RP into a close returns map (`convert_recurrence_matrix()`)
- of this "rotated" RP get the horizontal lines and its starting indices (`horizontalhisto()`)
- encode the line lengths into the close returns map (`create_close_returns_map()`)
- iterate over all lines, starting from the longest one, and delete all neighboring lines recursively (`get_final_line_matrix()`)
- from the "left-over"-lines built the final returned skeletonized RP (`build_skeletonized_RP()`)

In the beginning I thought the recursive neighbor-line-searching in `get_final_line_matrix` would be the bottle-neck. Turns out that this is not the case. In fact, the initial search for all lines and their starting indices in `horizontalhisto()` is the most time consuming part:

Bildschirmfoto 2021-12-14 um 11 49 01

Follwing on this and your suggestions @heliosdrm in #98 I rewrote `_linehistograms` to serve the purpose. And truly that idea needs less allocations, but nevertheless it has longer runtime, the larger the input RP grows:

computation_time_line_histogram

So finally I sticked to my initial implementation. I guess, if one really would want it one could work on vertical instead of horizontal lines, making use of Julia's column major thingy, or one operates on the diagonal lines directly. The latter would include to rewrite the "line-neighboring"-definition and -search. I do not know if it is worth the trouble to be honest. What do you think?

Any suggestions are welcome.

Best,

Hauke


hkraemer:
Okay @heliosdrm , I altered the `runtests.jl`-file and make it download some time series from `https://raw.githubusercontent.com/JuliaDynamics/JuliaDynamics/master/timeseries`. @Datseris and I agreed on this in `DelayEmbeddings.jl`, since we encountered so diverse trajectories when integrating chaotic systems on different OS' while testing. This ensures controlability and comprehensible tests. Hope you are alright with this?
Best

Project.toml
```diff
@@ -4,11 +4,14 @@ repo = "https://github.com/JuliaDynamics/RecurrenceAnalysis.jl.git"
...
 
 [deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 DelayEmbeddings = "5732040d-69e3-5649-938a-b6b4f237613f"
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+DynamicalSystemsBase = "6e36e845-645a-534a-86f2-f5d4aa5a06b4"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
```
Datseris, line 14: None of these three have a reason to be dependencies. Especially PyPlot.

Datseris (approved):
@hkraemer What's going on with the dependencies here?

Project.toml
hkraemer: Forgot about that one. Good point, thank you. Removed it.

src/matrices/skeletonization.jl
```diff
@@ -0,0 +1,396 @@
...
+# Core function, which gets exported
+"""
+    skeletonize(R) → R_skel
+
+Skeletonizes the recurrence matrix `R` (see [`recurrence_matrix`](@ref)) by using
+the algorithm proposed by Kraemer & Marwan [^Kraemer2019]. This function returns
+`R_skel`, a recurrence matrix, which only consists of diagonal lines of
+"thickness" one.
+
+## References
```
Datseris, line 15: You don't need the header `## References` because the references go to the end of the documentation once compiled.

src/matrices/skeletonization.jl
```diff
@@ -0,0 +1,396 @@
+#=
+In this file the core computations for skeletonizing a recurrence matrix
+are defined.
+=#
+
+# Core function, which gets exported
+"""
+    skeletonize(R) → R_skel
+
+Skeletonizes the recurrence matrix `R` (see [`recurrence_matrix`](@ref)) by using
```
Datseris, line 10: Notice that the function `recurrence_matrix` is not part of the Public API, so it is not in the docs. This link will not be valid. You can use `RecurrenceMatrix` instead, but I think no need to link anything here.

src/matrices/skeletonization.jl
```diff
@@ -0,0 +1,396 @@
...
+        # numbers, equal to their lengths
+        X_hori1 = create_close_returns_map(lines_copy1t, lines_copy1l, lines_copy1c, size(X_cl1))
+        X_hori2 = create_close_returns_map(lines_copy2t, lines_copy2l, lines_copy2c, size(X_cl2))
+    end
+
+    # scan the lines, start with the longest one and discard all adjacent lines
+    lines_final_t, lines_final_l, lines_final_c = get_final_line_matrix(lines1t, lines1l, lines1c, lines_copy1t, lines_copy1l, lines_copy1c, X_hori1)
+
+    # if not symmetric input RP, than compute for the upper triangle as well
+    if ~symm
```
Datseris, line 59:
```suggestion
    if !symm
```

The `!` is the typical operator used for "not".

test/runtests.jl
```diff
@@ -1,6 +1,23 @@
 using RecurrenceAnalysis
 using Test
 
+# Download some test timeseries
+# Old:
+repo = "https://raw.githubusercontent.com/JuliaDynamics/NonlinearDynamicsTextbook/master/exercise_data"
+tsfolder = joinpath(@__DIR__, "timeseries")
+todownload = ["$n.csv" for n in 1:4]
```
Datseris, line 8: Maybe we should put these test in the JuliaDynamics/JuliaDynamics repo? The same repo that hosts videos and/or images for documentation.

test/smallmatrix.jl
```diff
@@ -121,7 +121,54 @@ rmat = CrossRecurrenceMatrix(sparse(i,j,trues(length(i))))
     end
 end
 
-### Recurrence network
+### Skeletonization of RPs
```
Datseris, line 124: Could you please put these tests in a new test file?

test/smallmatrix.jl
```diff
@@ -121,7 +121,54 @@ rmat = CrossRecurrenceMatrix(sparse(i,j,trues(length(i))))
...
+@testset "Skeletonized recurrence structures" begin
+    data = sin.(2*π .* (0:1000)./ 60)
```
Datseris, line 126: Here it is useful to start a 2nd test set named "sinusoidal"

test/smallmatrix.jl
```diff
@@ -121,7 +121,54 @@ rmat = CrossRecurrenceMatrix(sparse(i,j,trues(length(i))))
...
+
+    @test tauRR2 == tauRR3
+
+    c1 = findall(x->x!=0, diff(tauRR1))
+    c2 = findall(x->x!=0, diff(tauRR2))
+
+    @test mean(diff(c1[3:2:end])) == 60
+    @test mean(diff(c2[3:2:end])) == 60
+
+    data = readdlm(joinpath(tsfolder, "test_time_series_lorenz_standard_N_10000_multivariate.csv"))
```
Datseris, line 151: Here start a second test set named "chaotic"

Datseris (commented):
It is probably a good idea to also open a PR in DynamicalSystems.jl adding this to the docs. We could make a simple plot where the left is the normal RP and the right plot is the skeletonized

test/runtests.jl
hkraemer: You want to put the tests into that repo? Why?

test/smallmatrix.jl
hkraemer: done

test/smallmatrix.jl
hkraemer: done

test/smallmatrix.jl
hkraemer: done

src/matrices/skeletonization.jl
hkraemer: done

hkraemer:
I will also make a PR to DynamicalSystems with an appropriate Doc-entry.

test/runtests.jl
Datseris: Because this repo is associated with our book project. Its contents might change in the second edition. Speculative stuff.

hkraemer: I see. Pretty speculative stuff, indeed. Okay, now I get your point: You want to migrate the timeseries stored in "https://raw.githubusercontent.com/JuliaDynamics/NonlinearDynamicsTextbook/master/exercise_data" into "https://raw.githubusercontent.com/JuliaDynamics/JuliaDynamics/master/timeseries"? That makes sense. I was only confused because I thought you mean moving the actual tests into that repo. Am I getting this right now? Then we also have to change tests in `DelayEmbeddings.jl`, I can make a new PR for all that separately.

Datseris: Hahah yes sorry, I only meant to transfer the .csv files to the second link that you posted!

Datseris:
I'll wait a bit on whether @heliosdrm can give a review here, otherwise will merge in a week or so.

hkraemer:
That's fine. I have already prepared a PR at DynamicalSystems.jl, which I will push, when this one here is merged.

Datseris:
Just open the PR already, its fine even if it doesn't build on CI I can still read it.

heliosdrm:
Hi. I need a bit of spare time to look into this - bad dates. :disappointed: I hope having time in the Christmas break. If it can wait until then... Otherwise go ahead. Thank you.

Datseris:
@hkraemer I'd say we wait, I don't see any pressing need to merge ASAP. BTW, whenever you open a PR that adds a new feature don't forget to do the following:
* increment minor version number in Project.toml
* Add a changelog entry

heliosdrm:
The skeletonization procedure is new to me, so I don't feel really qualified to judge the details. I'm sure you know well the good way to implement the algorithm published in your own paper! :wink:

If anything, I would say that `scan_lines!` might look nicer if, instead of using multiple `ifelse` block, there was a single recursive call to `scan_lines!` where the last two arguments are calculated in another function of `li`, `co`, `N`, `M` and `i`. But that's mostly a cosmetic issue, so it could be left for later on, and now merge it as it is.

Regarding the files that are downloaded in the tests: I'm not sure of having understood the point completely, but if they are potentially useful for more than one package, what do you (and @Datseris ) think of bundling them as an artifact instead?

Datseris:
> Regarding the files that are downloaded in the tests: I'm not sure of having understood the point completely, but if they are potentially useful for more than one package, what do you (and @Datseris ) think of bundling them as an artifact instead?

The only use for these files is in the test suite. Artifact's purpose is more user oriented, not developer oriented. Just downloading them from a central repo is simpler than making an artifact I think.

---

Indeed I agree that the source code can be made clearer and with much less duplication. Seems like every function is composed of large `if` clauses, and some parts of the code clearly repeat in each clause (e.g., each close has `x = size(R)` and other things.

src/matrices/skeletonization.jl
```diff
@@ -0,0 +1,394 @@
...
+    liness = liness[:, no_zero_lines]
+    ls = size(liness)
+    liness = liness[:,sortperm(@view liness[1, :]; rev=true)]
+    return vec(liness[1,:]), vec(liness[2,:]), vec(liness[3,:])
+end
+
+# function horizontalhisto2(R::SparseMatrixCSC)
+#     # Transpose R in order to get the "horizontal" lines
+#     # rows = colvals(R[2:end,:])
+#     # cols = rowvals(R[2:end,:])
```
Datseris, line 282: What's the purpose of this huge block of commented code?

heliosdrm:
Since it is a lot of new code, if it is working fine,  perhaps we could merge it to make it part of the repo history, and create a new issue to clean `skeletonization.jl`.

src/matrices/skeletonization.jl
hkraemer: This function serves the same purpose as `horizontalhisto()` does. In contrast to the latter, in this approach I attempted to make it faster than `horizontalhisto()`. This block is so huge, because it actually copies `_linehistograms` a lot, which is the very performant core-function for RQA by @heliosdrm . After our little discussion in #98 I thought it might be useful to reuse that code and compute the rotated lines in the skeletonization method. It turned out that it is in fact slower for larger input (see my posted figure above). I left it for @heliosdrm , maybe he might be able to make it work faster. For the actual implementation it is not necessary.

Regarding your clean-code-isssues, @Datseris and @heliosdrm : I am very thankful for any suggestions to make it cleaner and look nicer. Just let me know or simply do it, if it is obvious. I do not care personally if we merge it now or later after cleaning.

heliosdrm: Ok, I'll give a closer look to the two versions of horizontal histograms and see if I can find some improvement. Would you be so kind to provide a script to reproduce the benchmarking above?

I'll also propose "cleaner" versions of the code.

Datseris:
Personally I'd prefer cleaning the code in this PR.

src/matrices/skeletonization.jl
hkraemer: Sure! Try something like this AFTER outcommenting `horizontalhisto2()`:
```
using RecurrenceAnalysis
using LinearAlgebra
using TimerOutputs

using PyPlot
pygui(true)

data = sin.(2*π .* (0:5500)./ 60)
Y = embed(data, 3, 15)

timings_my_way = zeros(19)
timings_based_on_linehistograms = zeros(19)
for (i,N) = enumerate(500:250:5000)
        println("i is: $i")
        println(N)
        X = RecurrenceMatrix(Y[1:N,:], 0.25; fixedrate=true)
        X_cl1 = RecurrenceAnalysis.convert_recurrence_matrix(tril(X))
        timings_my_way[i] = @elapsed RecurrenceAnalysis.horizontalhisto(X_cl1)
        timings_based_on_linehistograms[i] = @elapsed RecurrenceAnalysis.horizontalhisto2(X_cl1)
end


fig = figure(figsize = (10,5))

plot(vec(500:250:5000), timings_my_way, color="red", label="actual method")
plot(vec(500:250:5000), timings_based_on_linehistograms, color="blue", label="method based on _linehistograms")
ylabel("computation time")
xlabel("trajectory length")
grid()
legend()
```

Thanks @heliosdrm for taking a closer look. Also looking forward to your suggested ways of eliminating the ugly `ifelse`-blocks, @Datseris and @heliosdrm .

heliosdrm:
Mm. My benchmark doesn't look as @hkraemer 's:
Figure_1

```
julia> versioninfo()
Julia Version 1.6.1
Commit 6aaedecc44 (2021-04-23 05:59 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i3-4100M CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, haswell)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 4
```

(Oh, by the way, I see that I have to update my Julia version!)

hkraemer:
This is weired...

heliosdrm:
Same result with Julia 1.7 - except that both versions are about 1 s faster at N = 5000; but again the "new histo" beats the "actual method" at large values of N
 
```
julia> versioninfo()
Julia Version 1.7.0
Commit 3bf9d17731 (2021-11-30 12:12 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i3-4100M CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, haswell)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 4
```

heliosdrm:
Today I tried it in a Windows machine, and I could reproduce your results (the times I got were two times the ones you reported, but they followed the same trend). It looks as if the kind of operations that are being done with the matrices perform differently in either OS. This is the configuration I tried:

```
julia> versioninfo()
Julia Version 1.7.0
Commit 3bf9d17731 (2021-11-30 12:12 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, broadwell)
Environment:
  JULIA_DEPOT_PATH = D:\.julia
  JULIA_EDITOR = "C:/Program Files/Microsoft VS Code/bin/code.cmd"
  JULIA_NUM_THREADS = 4
```

heliosdrm:
I don't know if this is relevant for the Windows vs. Linux differences, but anyway...

I have seen that the strategy used to make the histograms consists in building a matrix with three rows, which is eventually splitted, and the three rows are returned separately. That matrix has initially one single column, and it is progressively replaced by matrices with more columns, extending them with `hcat`. At the end of the function, the matrix is replaced by another with less columns that is obtained by subindexing the larger one.

I don't know the implementation details of Julia to be sure of this, but I have the gut feeling that this is not optimal. Matrices can't be extended or reduced "in-place", but vectors can. So, probably things might go better if instead of using a three-rowed matrix, those functions work directly with three vectors (extending them by `append!` or `push!`, and finally reducing them with `delete!`).

Also, regarding this comment:

> I  guess, if one really would want it one could work on vertical instead of horizontal lines, making use of Julia's column major thingy, or one operates on the diagonal lines directly. The latter would include to rewrite the "line-neighboring"-definition and -search. I do not know if it is worth the trouble to be honest. What do you think?

I'm not sure about operating directly on the diagonal lines, but I'm pretty confident that working with column-based indexing instead of row-based should have a positive impact on performance (and would not be too complicated, I hope).


Besides, although I have only looked into those functions very superficially, it looks that the reason of subindexing the matrix

hkraemer:
> I don't know the implementation details of Julia to be sure of this, but I have the gut feeling that this is not optimal. Matrices can't be extended or reduced "in-place", but vectors can. So, probably things might go better if instead of using a three-rowed matrix, those functions work directly with three vectors (extending them by append! or push!, and finally reducing them with delete!).

I tried this and it did not improve on computation time. - At least on my Mac-OS-machine. I find it strange that the different OS' have such different computation times... But this is too much IT-stuff I am not familiar with. 

I can try the column-lines-approach in the first week of January.

Datseris:
> I have seen that the strategy used to make the histograms consists in building a matrix with three rows, which is eventually splitted, and the three rows are returned separately. That matrix has initially one single column, and it is progressively replaced by matrices with more columns, extending them with hcat. At the end of the function, the matrix is replaced by another with less columns that is obtained by subindexing the larger one.

That seems like a bad idea. A better alternative is using a `Vector{Vector}` instead exactly as Helios mentions afterwards! 

> but I'm pretty confident that working with column-based indexing instead of row-based should have a positive impact on performance 

Oh yes, this is certainly guaranteed, I used to teach it in my Julia courses. But keep in mind that its impact on _sparse arrays_ will not be that much, because the internal storage of sparse arrays is not matrices in reality.

Datseris:
@hkraemer 

> I tried this and it did not improve on computation time. - At least on my Mac-OS-machine.

Even if this is true, the code with `Vector{Vector}` is much more intuitive Julia code. Matrices are really not supposed to have their number of columns altered like that. It will make the source code simpler to reason for and about, and as such identify performance bottlenecks better. 

For now, the best move forwards is a perforamance estimation of which lines of the code are worse. If you are using VSCode, it was a nice way to do this with `@profile` macro. Otherwise, you can use flame graphs from here: https://timholy.github.io/FlameGraphs.jl/stable/ 

Have you also checked the code for type stability with the macro `@code_warntype` ?

src/matrices/skeletonization.jl
```diff
@@ -69,62 +69,29 @@ end
 
 # Transforms the standard RP into a close returns map
 function convert_recurrence_matrix(R::SparseMatrixCSC; triangle::Bool = true)
```
heliosdrm, line 71: The name `convert_recucrrence_matrix` is ambiguous. Since this is the counterpart of `revert_close_return_map`, I think that it would be better to reuse the name `create_close_returns_map` that is given to another function below. It would not be ambiguous, because the number of required arguments is different, so they would be two different methods.

Also, I have left `triangle` as a keyword argument to avoid breaking other parts of the code, but I see no strong reason to keep it as such. There are no more optional arguments, so it could be a "normal" positional argument.a keyword a

heliosdrm: Oh, I noticed that the "simplified" versions of the functions I have modified in this commit are not fully equivalent to the original ones written by @hkraemer : I have assumed that both the inputs and outputs are sparse _Boolean_ matrices, and that all nonzeros are `true`. On the other hand, the original implementation made "pseudo-rotations" of matrices with any type of elements.

That's very easy to fix: just take `nonzeros(R)` and pass its values as the third argument of `sparse` in the `return` lines, instead of a new vector made with `trues`. But I wonder if that's needed, or if my assumption was right and we can leave it as is.

src/matrices/skeletonization.jl
```diff
@@ -98,17 +98,20 @@ end
...
-function create_close_returns_map(lines1::Vector{Int}, lines2::Vector{Int}, lines3::Vector{Int}, size_of_cl_ret_RP::Tuple{Int, Int})
-    X = zeros(Int, size_of_cl_ret_RP)
-    for i = 1:length(lines1)
-        line_length = lines1[i]
-        line_ind = lines2[i]
-        column_ind = lines3[i]
-        for j = 0:line_length-1
-            X[line_ind,column_ind+j] = line_length
-        end
+function create_close_returns_map(lines1, lines2, lines3, dims)
```
heliosdrm, line 101: The code of this function was reasonably readable and compact, but it created a full matrix to eventually convert it to a sparse one. I have modified it to make directly a sparse matrix, which should perform better when the matrix has many zeros.

I have also removed the type annotations of `create_close_returns_map` - which are not really necessary right now.

heliosdrm:
More information about the difference in performance between the two functions to calculate the "horizontal histograms". The recurrence plots used in the previous comparisons were quite dense (RR= 0.25), so the advantage of using sparse arrays was not that much. Reducing the threshold to 0.0025 (and taking the average of ten repetitions, because with that smaller threshold the calculations are also faster and there is more noise in the elapsed times), I got the following times in Linux and Windows:

Figure 2021-12-30 083252

The "absolute" better performance of Linux vs. Windows is not relevant here, because they are different machines. The point is that, according to these results, with more sparse matrices the "alternative version" of the function is faster in both OS.

To be more precise, it seems that the alternative version is faster in both OS for small matrices, and that for larger matrices the advantage increases in Linux but decreases in Windows (as we had observed before); but the size at which the alternative is no longer faster in Windows depends on the sparsity of the RP (the sparser the better).

All in all, I think that the alternative version is more advantageous, since it always beats the current implementation in Linux, and even in other OS, recurrence plots should be rather sparse at the outset, isn't it?

hkraemer:
I find this convincing. Still, it is strange that on my machine it seems to be mirrored. Anyway, let's stick with the `_linehistograms`-version I suggest.

Datseris:
So what's the source version now? Is this good to be merged @heliosdrm ?

heliosdrm:
> So what's the source version now? Is this good to be merged @heliosdrm ?

Hi, happy new year! No, I think it still needs a bit of revision. I understood that @hkraemer would rewrite it to make column-first iterations. I have been examining the code and I think I can help during next week, if he's stuck or does not have time.

Datseris:
haha happy new year :P

hkraemer:
yepp, I'll go for it this week. Happy new year to you as well ;)

hkraemer:
Okay @heliosdrm and @Datseris , I implemented the whole thing based a close returns map which now has been rotated (so to say), in order to make use of the column-major feature of Julia.

Actually everything just switched from being horizontal to vertical (little bit tricky with all the indices etc., but it works now). Here is a little performance comparison:

This is the old version from 7c52049 , i.e. everything based on horizontal lines and my original method for creating the linehistograms:
performance_based_on_RP_lines

And this is the performance, when the computation is based on columns, i.e. vertical structures, still using my original implementation for creating the linehistograms:
performance_based_on_RP_columns

The first line `horizonal` is actually the call of `horizontalhisto` in the upper figure and `verticalhisto` in the lower picture.
As you can see, the approach based on vertical structures is faster now, still having the same allocations.

I also tried to use the `_linehistograms`-approach here, which we were discussing @heliosdrm , consequently in this commit it is named `verticalhisto2`. With this it looks like this:
performance_based_on_RP_columns_linehisto

It is slower than my initial implementation, the code is twice as long and the allocations needed are 5-6 times higher. So again, maybe that is my machine (MacBookPro 2015, 2,8 Ghz quad i7, 16GB 1600 DDR RAM) and on any other OS the implementation based on `_linehistograms`, i.e. `verticalhisto2`, is faster!? But this would be very, very strange then and I would be more than surprised. @heliosdrm , if you don't mind try this on your machines: You only have to comment `verticalhisto`, outcomment `verticalhisto2` and name it `verticalhisto`. 

The actual problem the figures above are based on was a `5000-by-5000`-Recurrence Plot of a chaotic Lorenz-trajectory.

Best

hkraemer:
Oh and concerning type-stability, @Datseris : All versions are type-stable.

And another small addition to my last post: When using Benchmarktools and `@btime`, my above statements do still hold, but the difference from my original implementation of `verticalhisto` and the one based on `_linehistograms` is smaller. Especially, the number of allocations are kind of similar that way with the memory twice as high (compared to my posted 5-6 time increase).  Still faster though, and when this holds also for different machines, I suggest we should stick to my original implementation. But maybe I messed something up in the actual implementation of `verticalhisto2` @heliosdrm ?
Best

heliosdrm:
In Linux the difference is still the opposite, with an overwhelming difference!

I used this code (taking the Lorentz trajectory downloaded from the tests):

```julia
using RecurrenceAnalysis
using TimerOutputs, BenchmarkTools
to = TimerOutput()
data = readdlm(joinpath(tsfolder, "test_time_series_lorenz_standard_N_10000_multivariate.csv"))
X = RecurrenceMatrix(data[5001:10000,:] ,0.01, fixedrate=true)
X_cl1 = RecurrenceAnalysis.create_close_returns_map(tril(X))
# First repetition to avoid JIT time
RecurrenceAnalysis.verticalhisto(X_cl1);
RecurrenceAnalysis.verticalhisto2(X_cl1)
@timeit to "current version" RecurrenceAnalysis.verticalhisto(X_cl1)
@timeit to "alternative version" RecurrenceAnalysis.verticalhisto2(X_cl1)
```
```julia-repl
julia> to
 ──────────────────────────────────────────────────────────────────────────────
                                       Time                   Allocations      
                               ──────────────────────   ───────────────────────
       Tot / % measured:             327s / 0.28%           4.01GiB / 21.8%    

 Section               ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 current version           1    846ms  92.7%   846ms    696MiB  77.9%   696MiB
 alternative version        1   66.6ms  7.30%  66.6ms    197MiB  22.1%   197MiB
 ──────────────────────────────────────────────────────────────────────────────

julia> @benchmark RecurrenceAnalysis.verticalhisto(X_cl1)
BenchmarkTools.Trial: 6 samples with 1 evaluation.
 Range (min … max):  806.429 ms …    1.123 s  ┊ GC (min … max): 3.75% … 3.98%
 Time  (median):     867.303 ms               ┊ GC (median):    3.70%
 Time  (mean ± σ):   896.964 ms ± 116.832 ms  ┊ GC (mean ± σ):  3.81% ± 0.16%

  ██       █    █   █                                         █  
  ██▁▁▁▁▁▁▁█▁▁▁▁█▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  806 ms           Histogram: frequency by time          1.12 s <

 Memory estimate: 696.17 MiB, allocs estimate: 212594.

julia> @benchmark RecurrenceAnalysis.verticalhisto2(X_cl1)
BenchmarkTools.Trial: 164 samples with 1 evaluation.
 Range (min … max):  29.330 ms … 46.171 ms  ┊ GC (min … max): 14.87% … 16.21%
 Time  (median):     29.817 ms              ┊ GC (median):    14.82%
 Time  (mean ± σ):   30.589 ms ±  2.080 ms  ┊ GC (mean ± σ):  15.93% ±  1.98%

   █▆▅▁▁                                                       
  ▆█████▅▆▃▅▃▁▃▁▁▁▁▁▃▁▄▅▁▃▄▃▄▄▄▃▄▃▃▃▁▁▁▁▁▁▁▁▃▁▁▁▁▁▃▁▁▁▃▁▁▁▁▁▃ ▃
  29.3 ms         Histogram: frequency by time        35.7 ms <

 Memory estimate: 197.28 MiB, allocs estimate: 11643.
```

I'll try in Windows later on

hkraemer:
Thanks for that. This is so super duper strange, I don't get it and I'm afraid that I am not able to understand why that is.

Datseris:
@hkraemer have you used exactly the same code as @heliosdrm? Without any commenting/uncommenting business, as this can sometimes lead to wrong ordering of code execution. Can you please paste exactly the output of this code snippet as Helios has done?

heliosdrm:
I got a similar result in Windows (same code as above):

```julia-repl
julia> to
 ──────────────────────────────────────────────────────────────────────────────
                                       Time                   Allocations
                               ──────────────────────   ───────────────────────
       Tot / % measured:            59.7s / 2.47%           3.32GiB / 26.3%

 Section               ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 current version            1    1.32s  89.6%   1.32s    696MiB  77.9%   696MiB
 alternative version        1    154ms  10.4%   154ms    197MiB  22.1%   197MiB
 ──────────────────────────────────────────────────────────────────────────────

julia> @benchmark RecurrenceAnalysis.verticalhisto($X_cl1)
BenchmarkTools.Trial: 5 samples with 1 evaluation.
 Range (min … max):  1.047 s …   1.081 s  ┊ GC (min … max): 5.46% … 5.47%
 Time  (median):     1.059 s              ┊ GC (median):    5.59%
 Time  (mean ± σ):   1.061 s ± 12.963 ms  ┊ GC (mean ± σ):  5.59% ± 0.16%

  █          █        █            █                      █
  █▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.05 s         Histogram: frequency by time        1.08 s <

 Memory estimate: 696.17 MiB, allocs estimate: 212594.

julia> @benchmark RecurrenceAnalysis.verticalhisto2($X_cl1)
BenchmarkTools.Trial: 64 samples with 1 evaluation.
 Range (min … max):  64.498 ms … 139.122 ms  ┊ GC (min … max): 13.95% … 17.15%
 Time  (median):     77.788 ms               ┊ GC (median):    15.09%
 Time  (mean ± σ):   78.959 ms ±  10.347 ms  ┊ GC (mean ± σ):  15.18% ±  2.13%

       ▃ ▃       ███ ██ ▃ ▃▃███    █ ▃      ▃  ▃
  ▇▁▁▁▇█▁█▇▇▇▇▇▇▇███▇██▁█▇█████▇▇▇▇█▇█▁▁▁▇▁▁█▁▁█▁▁▁▁▇▁▁▁▁▁▁▇▁▇ ▁
  64.5 ms         Histogram: frequency by time         97.9 ms <

 Memory estimate: 197.28 MiB, allocs estimate: 11642.
```

Datseris:
Is it simple to allow an option via a keywrod argument? We can simply say: If you expect to have a lot of lines, use option A, otherwise use option B. Since we have the code for both options anyways, maybe it isn't that hard to support. This also resolves the difficult decision of choosing one option to put in the released version.

hkraemer:
Jepp, I am fine with that. @heliosdrm , you decide which method will be the default =)

heliosdrm:
In Linux, with fixed RR=8% both versions take more or less the same time, although in terms of memory allocation they look more like @hkraemer 's:

```
julia> to
 ──────────────────────────────────────────────────────────────────────────────
                                       Time                   Allocations      
                               ──────────────────────   ───────────────────────
       Tot / % measured:            12.5s / 25.2%           12.2GiB / 100%     

 Section               ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 current version            1    1.78s  56.5%   1.78s   1.69GiB  13.9%  1.69GiB
 alternative version        1    1.37s  43.5%   1.37s   10.5GiB  86.1%  10.5GiB
 ──────────────────────────────────────────────────────────────────────────────
```

I'm not sure which should be default. What situation is more typical for applying the skeletonization algorithm? This RP (RR=1%)?

Figure_1-1


Or this (RR=8%)?
Figure_1-2

hkraemer:
I'd say higher thresholds, that is why I am digging in on this. Especially when dealing with noisy data. The skeletonization can also serve as a noise-filter and estimator (described in the paper). But actually this whole discussion is pretty academic ;) I guess nobody cares in the end, especially because the number of lines only increases rapidly when using chaotic dynamics. For regular dynamics the number of lines even with higher thresholds might not be as high. Long story short: I vote for the current version as default.

heliosdrm:
I think that we can now settle the discussion of what version of `verticalhisto` should be used. See the extended benchmark with the "updated version" just introduced in 79fb20a :smiley: 

```
julia> to
 ──────────────────────────────────────────────────────────────────────────────
                                       Time                   Allocations      
                               ──────────────────────   ───────────────────────
       Tot / % measured:            5118s / 0.07%            265GiB / 4.76%    

 Section               ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 current version            1    1.78s  49.6%   1.78s   1.69GiB  13.4%  1.69GiB
 alternative version        1    1.37s  38.2%   1.37s   10.5GiB  83.2%  10.5GiB
 updated version            1    435ms  12.1%   435ms    440MiB  3.40%   440MiB
 ──────────────────────────────────────────────────────────────────────────────
```

heliosdrm:
Or well... let's wait, because the optimizations introduced in 79fb20a can also be done to the "alternative version". I still have to work on it.

Datseris:
Oh wow, we were still using the matrix version instead of individual vectors...? I thought that was the first thing we said to change :D 

This is also not good:
```
Rjdiffs = diff([0; @view R[:,j]; 0])
```
Why create the intermediate array `[0; @view R[:,j]; 0]`? Seems to me that we can write a simple 5 lines of code function that calculates this difference without extra allocations...

heliosdrm:
After optimizing both versions, the "alternative" version `verticalhisto2` again beats the "current one" in Linux. This is the benchmark with the same RPs of the Lorenz trajectory  discussed in previous commentaries (tested with both fixed RR=0.08, and RR=0.01, see code above), and the versions of the functions in commit 37df5d2:

@hkraemer , can you test it in MacOS?

```
 ──────────────────────────────────────────────────────────────────────────────────────
                                                Time                   Allocations      
                                        ──────────────────────   ──────────────────────
                Tot / % measured:           90.9s / 0.65%            456MiB / 26.4%    

 Section                       ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────────────
 current version RR=0.08            1    385ms  65.0%   385ms   55.9MiB  46.5%  55.9MiB
 current version RR=0.01            1    189ms  31.9%   189ms   51.9MiB  43.2%  51.9MiB
 alternative version RR=0.08        1   16.1ms  2.71%  16.1ms   11.0MiB  9.13%  11.0MiB
 alternative version RR=0.01        1   2.12ms  0.36%  2.12ms   1.47MiB  1.23%  1.47MiB
 ──────────────────────────────────────────────────────────────────────────────────────
```

hkraemer:
Cool, I will do tonight. Thanks a lot, looks like a huge improvement!

hkraemer:
Hey @heliosdrm , sorry for the delay. Looks good here as well! 

```
 ────────────────────────────────────────────────────────────────────────────────────────────────
                                                         Time                   Allocations      
                                                 ──────────────────────   ───────────────────────
                Tot / % measured:                      401s / 0.14%           15.3GiB / 0.77%    

 Section                                 ncalls     time   %tot     avg     alloc   %tot      avg
 ────────────────────────────────────────────────────────────────────────────────────────────────
 optimized current version RR=0.08            1    419ms  73.9%   419ms   55.9MiB  46.5%  55.9MiB
 optimized current version RR=0.01            1    137ms  24.2%   137ms   51.9MiB  43.2%  51.9MiB
 optimized alternative version RR=0.08        1   9.42ms  1.66%  9.42ms   11.0MiB  9.13%  11.0MiB
 optimized alternative version RR=0.01        1   1.23ms  0.22%  1.23ms   1.47MiB  1.23%  1.47MiB
 ────────────────────────────────────────────────────────────────────────────────────────────────
```

Datseris:
So, current version has no reason to exist, and will be replaced entirely by the alternative. Can you please do the change, cleanup the code, and then I give a final review before merging?

Datseris:
bump here

hkraemer:
Sorry @Datseris for the delay. I polished a bit and tested again. Should be okay now and maybe you find some more allocations to avoid :p

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # RecurrenceAnalysis.jl News
 
+## v1.6.3
```
Datseris, line 3:
```suggestion
## v1.7.0
```

test/runtests.jl
Datseris: @hkraemer you have already moved the timeseries to the other repo, right? Then we have to change the link above,

Datseris (approved):
Only some minor comments left that I have forgotten to post, sorry!

hkraemer:
I just copied the timeseries into the code repo (/JuliaDynamics/timeseries).

Datseris:
Yes, but the link in the code is still targeting the `NonlinearDynamicsTextbook` repo!

hkraemer:
Shall I push in the master branch for this fix?

Datseris:
? Huh, no just do the necessary changes I've outlined in this Pull Request (skelotinization branch). If you scroll above you will see some comments left by me.

hkraemer:
Locally I'll get an error when running all tests for not having `DataStructures` in the dependencies, but there is (under `extras`)! I also tried `Pkg.resolve()`, but this did not change. Maybe that is only on my local machine.

hkraemer:
Okay, I found the error. It is all good here, I guess.

PR #129, opened 2021-12-29
Do not pirate `show` for sparse matrices

heliosdrm:
This is not convenient anymore, as Julia represents big sparse matrices with Braille patterns since v1.6 (https://github.com/JuliaLang/julia/pull/33821). Actually, perhaps we could even use that representation for `AbstractRecurrenceMatrix`.


heliosdrm:
Well, no: for `AbstractRecurrenceMatrix` the special representation of this package is still convenient, because it incorporates the inversion of the Y-axis, as usual for recurrence plots.

Datseris:
ok, but also increment patch version in Project.toml

Issue #130, opened 2022-02-17
Comparison of calculation speed


pucicu:
I did some comparison of the calculation speed for RPs and simple RQA. More details at <https://github.com/pucicu/RP_Speed_Test>. The Julia implementation (RecurrenceAnalysis.jl) is (surprising to me) not the fastest one.

![Computation speed for recurrence plots and recurrence quantification measures for the Rössler system.](https://raw.githubusercontent.com/pucicu/RP_Speed_Test/master/software_speed.svg "Computation speed")

Have fun! 😁

Norbert

Datseris:
Hi Norbert!

Thanks a lot for your post. Generally speaking I would be happy to see that there can be improvements to the library. However, are we sure that these benchmarks are accurate? I had a look at the repo you cite. 

* For Pyunicorn it seems to me that the function being benchmarked is: [`rqa_summary`](http://www.pik-potsdam.de/~donges/pyunicorn/api/timeseries/recurrence_plot.html?highlight=rqa_summary#pyunicorn.timeseries.recurrence_plot.RecurrencePlot.rqa_summary). The docstring states that it computes "the recurrence rate RR, the determinism DET, the average diagonal line length L and the laminarity LAM".
* For RecurrenceAnalysis.jl, the function being benchmarked is [`rqa`](https://juliadynamics.github.io/DynamicalSystems.jl/dev/rqa/quantification/#RecurrenceAnalysis.rqa) which computes: 
![image](https://user-images.githubusercontent.com/19669089/154577285-666f2353-259a-4950-9229-7fcc3c19e695.png)
  So, it computes at least triple the amount of quantities as `rqa_summary`, and the quantities computed are also much more complex than average rates.

It comes to no surprise for me that our version is slower.

Furthermore, the benchmark also mixes the time it takes to compute the recurrence matrix with the time it takes to compute RQA measures. It would be more useful to separate those two, and would also help each library see where they can improve.

heliosdrm:
> For Pyunicorn it seems to me that the function being benchmarked is: [`rqa_summary`](http://www.pik-potsdam.de/~donges/pyunicorn/api/timeseries/recurrence_plot.html?highlight=rqa_summary#pyunicorn.timeseries.recurrence_plot.RecurrencePlot.rqa_summary). The docstring states that it computes "the recurrence rate RR, the determinism DET, the average diagonal line length L and the laminarity LAM".

[...]

PR #132, opened 2022-05-17
use `eachindex` and `axes` in loops

heliosdrm:
Replace syntax `for i in 1:length(x)` by `for i in eachindex(x)`, and `for i in 1:size(x,d)` by `for i in axes(x,d)`, as recommended.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # RecurrenceAnalysis.jl News
 
+## v1.8.1
+- Better support for iteration over `AbstractRecurrenceMatrix` with functions from `Base`.
+
 ## v1.8.0
 - Switch to Graphs.jl from LightGraphs.jl for RNA.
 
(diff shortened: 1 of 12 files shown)


Datseris:
As recommended by whom and why? :D I don't think it makes much of a difference but sure. 

What I've noticed, that most likely will make a lot of difference, is that we do
```

    for j in eachindex(y), i in eachindex(x)
        @inbounds d[i,j] = evaluate(metric, x[i], y[j])
    end
```
instead of using any of the `pairwise, rowwise, colwise` from Distances.jl, which are much more performant versions that use BLAS and SIMD acceleration to compute distances across many pairs of vectors.

Datseris:
I didn't know that this was an established consensus. Now that I do, I will try to follow it. Sure, this is a straight forward stylistic improvement.

> On the other hand, I don't understand this commentary

yeah, that was not really relevant with this PR, but I just saw it because it modified this source code. The short story is that Distances.jl recommends using `pairwise` when wanting to compute pairwise distances across collections of points. We don't do that here, and I believe because of that we miss some BLAS+SIMD acceleration that Distances.jl offers under the hood for some standard distances like Euclidean. It is a separate topic of course, so needs to be done in a different PR, but should definitely help with #130

PR #134, opened 2022-05-18
Update docs and source to be more harmonious with rest of DynamicalSystems.jl

Datseris:
To goal of this PR is to increase the quality of documentation and source code to reflect the standards of clarity that DynamicalSystems.jl strives for. This means clarifying docstrings more, using better variable names, removing duplication of source code, and making source code maintainance more natural, i.e., not allowing stuff like specifying distances by strings.

@heliosdrm I have a question, not related with documentation, but also just saw it: why is the default value for `parallel` false, and why is it that we even need a configuration for that? The parallel method should always be used if more than 1 threads are available...? What's the argument against that?

### DIFF
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -5,7 +5,6 @@ import Base.Meta.parse
 
 const DEFAULT_METRIC = Euclidean()
 
-
 export RecurrenceMatrix, CrossRecurrenceMatrix, JointRecurrenceMatrix,
        AbstractRecurrenceMatrix, WithinRange, NeighborNumber, FAN
 
@@ -31,7 +30,7 @@ export embed,
        sorteddistances,
        skeletonize,
        @windowed,
-       rna,
+       rna
 
 
 include("matrices/distance_matrix.jl")
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -1,7 +1,6 @@
 module RecurrenceAnalysis
 
 using Distances, Statistics, LinearAlgebra, SparseArrays, DelayEmbeddings, StaticArrays
-import Base.Meta.parse
 
 const DEFAULT_METRIC = Euclidean()
 
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -3,23 +3,7 @@ module RecurrenceAnalysis
 using Distances, Statistics, LinearAlgebra, SparseArrays, DelayEmbeddings, StaticArrays
 import Base.Meta.parse
 
-const METRICS = Dict(
-    "euclidean"=>Euclidean(),
-    "max"=>Chebyshev(),
-    "inf"=>Chebyshev(),
-    "cityblock"=>Cityblock(),
-    "manhattan"=>Cityblock(),
-    "taxicab"=>Cityblock(),
-    "min"=>Cityblock()
-)
 const DEFAULT_METRIC = Euclidean()
-getmetric(m::Metric) = m
-function getmetric(normtype::AbstractString)
-    normtype = lowercase(normtype)
-    !haskey(METRICS,normtype) && error("incorrect norm type. Accepted values are \""
-        *join(keys(METRICS),"\", \"", "\" or \"") * "\".")
-    METRICS[normtype]
-end
 
 
 export RecurrenceMatrix, CrossRecurrenceMatrix, JointRecurrenceMatrix,
--- a/src/RecurrenceAnalysis.jl
+++ b/src/RecurrenceAnalysis.jl
@@ -48,8 +48,7 @@ export embed,
        skeletonize,
        @windowed,
        rna,
-       # deprecated:
-       transitivity
+
 
 include("matrices/distance_matrix.jl")
 include("matrices/matrices.jl")
--- a/src/deprecate.jl
+++ b/src/deprecate.jl
@@ -9,7 +9,7 @@ for T in (:WithinRange, :NeighborNumber)
             @warn string("`", $call, "{", $T, "}(x::AbstractMatrix, y, ε; kwargs...)` is deprecated, use `", $call, "{", $T, "}(Dataset(x), y, ε; kwargs...)`")
             $call{$T}(Dataset(x), y, ε; kwargs...)
         end
-        
+
         @eval function $call{$T}(x, y::AbstractMatrix, ε; kwargs...)
             @warn string("`", $call, "{", $T, "}(x, y::AbstractMatrix, ε; kwargs...)` is deprecated, use `", $call, "{", $T, "}(x, Dataset(y), ε; kwargs...)`")
             $call{$T}(x, Dataset(y), ε; kwargs...)
@@ -81,3 +81,5 @@ function transitivity(R::AbstractRecurrenceMatrix)
     end
     trans = numerator / (sum(R²) - LinearAlgebra.tr(R²))
 end
+
+export transitivity
--- a/src/deprecate.jl
+++ b/src/deprecate.jl
@@ -85,7 +85,7 @@ end
 export transitivity
 
 
-
+import Base.Meta.parse
 const METRICS = Dict(
     "euclidean"=>Euclidean(),
     "max"=>Chebyshev(),
--- a/src/deprecate.jl
+++ b/src/deprecate.jl
@@ -83,3 +83,24 @@ function transitivity(R::AbstractRecurrenceMatrix)
 end
 
 export transitivity
+
+
+
+const METRICS = Dict(
+    "euclidean"=>Euclidean(),
+    "max"=>Chebyshev(),
+    "inf"=>Chebyshev(),
+    "cityblock"=>Cityblock(),
+    "manhattan"=>Cityblock(),
+    "taxicab"=>Cityblock(),
+    "min"=>Cityblock()
+)
+getmetric(m::Metric) = m
+function getmetric(normtype::AbstractString)
+    @warn "specifying metric with strings is deprecated! "*
+    "Use a formal instance of a `Metric` from Distances.jl, e.g., `Euclidean()`."
+    normtype = lowercase(normtype)
+    !haskey(METRICS,normtype) && error("incorrect norm type. Accepted values are \""
+        *join(keys(METRICS),"\", \"", "\" or \"") * "\".")
+    METRICS[normtype]
+end
--- a/src/matrices/matrices.jl
+++ b/src/matrices/matrices.jl
@@ -6,7 +6,7 @@ The low level interface is contained in the function
 `recurrence_matrix`, and this is where any specialization should happen.
 =#
 ################################################################################
-# AbstractRecurrenceMatrix type definitions and documentation strings
+# AbstractRecurrenceMatrix type hierarchy and extensions of methods
 ################################################################################
 const FAN = NeighborNumber
 export FAN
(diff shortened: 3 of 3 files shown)


Datseris:
Oh, also, I'll most certainly be deprecating here allowing strings to specify distances, i.e. `"euclidean"` instead of `Euclidean`. There is no reason any library should support this. We should leverage the incredible composability of Julia, where arbitrary metric spaces are represented by a `Metric`, which has meaning in code, rather than a `String` which has no meaning for a metric. It costs nothing to the user to use `Euclidean` instead of `"euclidean"` while it allows us to not care about, nor test, string inputs.

heliosdrm:
> Currently WIP, more changes will come.
> 
> @heliosdrm I have a question, not related with documentation, but also just saw it: why is the default value for `parallel` false, and why is it that we even need a configuration for that? The parallel method should always be used if more than 1 threads are available...? What's the argument against that?

It was @hkraemer who wrote that IIRC, but the fact is that the issue ***is*** related to the documentation: in the code of the functions, the default value of `parallel` is not  `false`, but the result of `length(x) > 500 && Threads.nthreads() > 1`. The logic is that `Threads.@threads` comes with a penalty, so if the input is short or there is only one thread, using that option is slower that a simple loop.

Datseris:
It was @asinghvi17 that did the parallelization PR in #90 . 

---

@hkraemer somethings is fundamentally wrong here. The documentation strings say that `RecurrenceMatrix{FAN}(x, ε)` is such that each point has `ε` neighbors. Hence, `ε::Int`. However, that's now what actually happens at the source code??? This `ε` is resolved at the function `get_fan_threshold`, which has `@assert 0 < ε < 1` ? This is completely inconsistent with what the documentation says.

pucicu:
although FAN means to have a fixed number of neighbours (i.e., INT) we use ε as a fraction with respect to the total number of states, i.e. ε = #neighbours / size(distance matrix, 1). therefore, it can only be 0 < ε < 1.

I hope the code is doing exactly this because I have used it in this way in previous analysis.

the doc strings should be modified accordingly, of course.

best
norbert

Datseris:
Oh, so it doesn't actually mean fixed number of neighbors, or fixed amount of neighbors. It means fixed RATIO of neighbors, which is something very different. FAN is missleading then and should be changed to something that conveys the information accurately without actually requiring more information. E.g., replace "amount" with "ratio" and replace "FAN" with "FRN".

Datseris:
I am also very confused by the line `@inbounds if evaluate(metric, x[i], y[j]) ≤ ( (ε isa Real) ? ε : ε[j] )` @heliosdrm do you know anything about this...? In which scenario is `ε` accessed by an integer? So different datapoints have different `ε`? I don't understand from the documentation strings when this could happen.

Datseris:
I admit, I am unhappy with this whole FAN situation. We should have had a much simpler and clearer design from the start... Oh well, always space to improve. 

To me, it seems that this "Fixed Amount of Neighbors" functionality is very similar in with the `fixedrate` functionality, that tries to achieve a fixed recurrence rate. Why is it that `fixedrate` is decided by a boolean keyword, and when true, it interprets `ε` as a fraction, while `FAN` needs to be a type parameter, that when given, it also interprets `ε` as a keyword? Either everything should be a type parameter, or everything should be a keyword argument. 

Perhaps it is best to define three new types `Distance(ε), RecurrenceRate(ε), NeighborRatio(ε)`, and pass these types in the place of `ε`. If no type is given but an `ε::Real` is given, the `Distance` version is used (as done now). Also, which of the keyword arguments are compatible with each other? is `scale` compatible with `NeigbhorRatio` or `RecurrenceRate`? if not, it should be an additional argument exclusively for the `Distance` version.

hkraemer:
> @hkraemer somethings is fundamentally wrong here. The documentation strings say that RecurrenceMatrix{FAN}(x, ε) is such that each point has ε neighbors. 

No, it does/did not:

> The parametrized constructor `RecurrenceMatrix{NeighborNumber}` creates the recurrence matrix
> with a fixed number of neighbors for each point in the phase space, i.e. the number
> of recurrences is the same for all columns of the recurrence matrix.
> In such case, `ε` is taken as the target fixed local recurrence rate,
> defined as a value between 0 and 1, and `scale` and `fixedrate` are ignored.

Actually, I find this docstring very precise and it tells the user exactly what the code in `get_fan_threshold` does. Where is the problem @Datseris ?

At the same time I am also unhappy with the special `FAN`-constructor-thingy, but that had been discussed before I implemented the `FAN`-method and I am still on the same position as I have been these days: It is your baby here @Datseris and @heliosdrm , so I think the main code-design decision should be made by you guys. Of course I would be happy to contribute and also to give an opinion. So my opinion is the following: 

1. Get rid of the `FAN`-constructor.
2. Until now `ε` is a `:<Real`-number, which directly translates into a distance, which is the used recurrence threshold for each point in state space. If the keyword `fixedrate=true` `ε`, it becomes a global recurrence rate and the actual used threshold for each point in state space is determined from the distribution of all pairwise distances. So, why not extending this to the `FAN`-method (or however you want to call this)? 
3. Similar to `fixedrate` find another keyword for `FAN` (e.g. `localfixedrate`), and if this keyword is used `ε` translates into the local recurrence rate (similar to the procedure when using `fixedrate`). For me as a user this would be the easiest way to work with it.

Actually, the nomenclature used here is a bit misleading I admit. The only difference of `FAN` to `fixedrate` is that the specified `ε` is a local AND global recurrence rate, so the actual recurrence threshold is different (we called it adaptive in the docstring and that's what it has been called in the literature) for each point in state space, but the number of neighbors is constant! So it is actually consistent with the name in my opinion. But instead of specifying the number of neighbors `k` the user specifies the corresponding local recurrence rate (which is nothing more than `k` divided by the total number of state space points) and the code will find the corresponding number of nearest neighbors used for that. This is way more user-friendly than an input of the amount `k` by the user! By the end of the day you want to work with an `RP` of known recurrence rate instead of trial and error many times with different `k` since you don't know the neighborhood-relations in your trajectory.

In order to not break the `fixedrate`-keyword, why not introducing `localfixedrate` as keyword for the `FAN` method used? I would then write the new docstring as clear as possible and focus on the difference of the two methods, if you want me to do it.

Cheers

pucicu:
I like the idea. But can we also keep the keyword “FAN”? else I would have to code that it is not working anymore and have to modify it again. (I’m a big fan of downwards compatibility)

hkraemer:
I personally have no problem with the keyword `FAN`, but from the comments above I suspect @Datseris may have an issue with this?

pucicu:
keeping both?
;-)

Datseris:
> Actually, I find this docstring very precise and it tells the user exactly what the code

Indeed, it tells exactly what is going on _in the second sentence of the docstring_. In the first sentence it gives the expectation that I would be able to specify this amount of neighbors, which I can't. I admit it was a mistake from my side to suggest back then to use the existing name `NeighborNumber`, because this name means "specify the number of neighbors". In any case, the FAN naming should go away because it is not precise enough (at least in my opinion). Something that explicitly mentions the word "rate" is more precise. 

---

@pucicu no worries about `FAN` staying: it will be properly deprecated, which means that when you use it you'll be getting a warning telling you how to modify the code, but the code will run as before. Be aware that deprecated code is typically removed later down the line in a major version increase so it's not a good idea to continue using it.

---

@hkraemer here is the reason I am against having all these multiple keyword arguments: in Julia, keyword arguments coexist and can potentially all be specified at the same time. It is very rare that keyword arguments are mutually exclusive and in my head I can only remember the `range` function having mutually exclusive keywords. Furthermore, I think keywords should not change algorithms, but rather tune arguments of algorithms. Here we have three different ways (as far as actual computations done) to compute the Recurrence Matrix. So, I think, it is more sensible, and syntactically more Julian, to have actual types given instead of `ε`. From Hauke's comment, it seems like the names `RecurrenceThreshold, GlobalRecurrenceRate, LocalRecurrenceRate` are the most fitting, and most accurately describing what each version does. Yay or Nay on these three names? Of course, if none type is given, but an `ε::Real` is given directly, `RecurrenceThreshold` version is assumed.

To be sure I understand what is going on, here is what I think the difference between `GlobalRecurrenceRate, LocalRecurrenceRate` is  (let's use `r in (0, 1)` for ratio from now on please, and reserve `ε::Real` for explicit distance):

- `GRR(r)` establishes a fixed number of recurrences, specifically ratio `r` versus the total possible number of recurrences, which are `N^2` if `N = length(dataset)`. This means that each column of the matrix does not necessarily have the same amount of recurrences, but the sum of columns are guaranteed to have the specified number regardless of input dataset.
- `LRR(r)` establishes that each column of the recurrence matrix will have a ratio of `r` recurrences, out of the possible `N`. Hence, each point will have `k = round(Int, r*N)` neighbors.

In either case, it is not guaranteed that the RM is symmetric, only in the `RecurrenceThreshold(ε::Real)` it is. Is this all correct? If not, please do say so. If yes, that's exactly what I'll write in the docstring.

---

Similarly with `FAN`, the existing keyword `fixedrate` will stay but be deprecated.

---

A final question: can someone please tell me how `scale` comes into all this? Is `scale` a valid option for all three cases? By reading the documentation string, it appears to me that scale is only valid for `RecurrenceThreshold(ε::Real)` version. If this is true, then `scale` should become an argument of the `RecurrenceThreshold` type (and given the default value `1`). Actually, I guess scale should be yet another actual type, because of `scale` is given, once again `ε` is interpreted as a ratio. So, we would have four methods with the extra `ScaledRecurrenceThreshold(r, scale)` (better names welcomed).

Datseris:
btw having types and defining behavior via multiple dispatch will make for more readable code as a side benefit.

pucicu:
> Am 19.05.2022 um 23:45 schrieb George Datseris ***@***.***>:
> @pucicu <https://github.com/pucicu> no worries about FAN staying: it will be properly deprecated, which means that when you use it you'll be getting a warning telling you how to modify the code, but the code will run as before. Be aware that deprecated code is typically removed later down the line in a major version increase so it's not a good idea to continue using it.
> 
but this worries me. i dont use it every day. next time i’m using it, several version jumps will have made my code non-working. it would not be the first time I experience this in Julia (and in Python).


> 
> To be sure I understand what is going on, here is what I think the difference between GlobalRecurrenceRate, LocalRecurrenceRate is (let's use r in (0, 1) for ratio from now on please, and reserve ε::Real for explicit distance):
> 
> GRR(r) establishes a fixed number of recurrences, specifically ratio r versus the total possible number of recurrences, which are N^2 if N = length(dataset). This means that each column of the matrix does not necessarily have the same amount of recurrences, but the sum of columns are guaranteed to have the specified number regardless of input dataset.
> LRR(r) establishes that each column of the recurrence matrix will have a ratio of r recurrences, out of the possible N. Hence, each point will have k = round(Int, r*N) neighbors.
> In either case, it is not guaranteed that the RM is symmetric, only in the RecurrenceThreshold(ε::Real) it is. Is this all correct? If not, please do say so. If yes, that's exactly what I'll write in the docstring.
> 
although your understanding on `GRR` and `LRR` is correct, the conclusion that the _RM_ for `GRR` could be asymmetric is not correct. for `GRR`, in principle a _RM_ using aa fixed threshold is calculated, but using such a choice that the desired _RR_ is guaranteed. for `RecurrenceThreshold` and `GRR`, the definition of a recurrence is the same. the difference is only how to select the threshold. `LRR` is a different definition of a recurrence: instead of using a spatial distance, a given number of neighbours is selected.

> 
> A final question: can someone please tell me how scale comes into all this? Is scale a valid option for all three cases? By reading the documentation string, it appears to me that scale is only valid for RecurrenceThreshold(ε::Real) version. If this is true, then scale should become an argument of the RecurrenceThreshold type (and given the default value 1). Actually, I guess scale should be yet another actual type, because of scale is given, once again ε is interpreted as a ratio. So, we would have four methods with the extra ScaledRecurrenceThreshold(r, scale) (better names welcomed).
> 
I never have used this but I suppose it was implemented by @heliosdrm because of the RQA software of Webber who has implemented a kind of rescaling of the distance matrix (as far as I remember). I personally would not do such. better to use normalised or rescaled time series if necessary.
https://tocsy.pik-potsdam.de/CRPtoolbox/?q=content_faq#webber <https://tocsy.pik-potsdam.de/CRPtoolbox/?q=content_faq#webber>

The scaling, indeed, only affects the `RecurrenceThreshold` option. but the term `ScaledRecurrenceThreshold` might mislead to the impression that the threshold is scaled (which, in fact, might give the same results but it is not what is technically done, I guess).

pucicu:
FAN does mean the general approach that we use a fixed number of neighbors. it is not meant to denote the value given by the variable epsilon. therefore FAN is correct. it is also convention in the literature. we should not change it.

PR #148, opened 2023-03-01
Documentation improvements for rt_entropy functions

### LINKED ISSUE #146: Some doubts and suggestions about "rt_entropy" metric in RQA
#24 
#34 

Hello, dear developers.
There are some doubts and suggestions here.
First of all, as mentioned above in this proposal, the Julia RQA package introduces a new indicator called recurrence time entropy, which is called "rt_entropy" in the latest package（see as https://juliadynamics.github.io/DynamicalSystems.jl/latest/rqa/quantification/#RecurrenceAnalysis.rt_entropy ）, and I read the documentation and found that there is no definition, formula, or any source of reference, can you add relevant source references to the document. After reading the case, I re-read the reference "Recurrence Quantification Analysis: Theory and Best Practices." However, as mentioned in section 1.3.3 of the book, three indicators are introduced:
(1) recurrence period density entropy,
(2) mean recurrence time，
(3) number of the most probable recurrence time，
but the indicator "recurrence time entropy" is not mentioned in the book.
The "recurrence period density entropy" metric is quoted from the paper 《Exploiting Nonlinear Recurrence and Fractal Scaling Properties for Voice Disorder Detection》. In the article, it is also called "recurrence probability density entropy". It is a normalized dimensionless metric with a range of [ 0,1]. This indicator is also called "recurrence period density entropy" on wikipedia. But in another article 《Recurrence threshold selection for obtaining robust recurrence characteristics in different embedding dimensions》, it is clearly stated that RTE is normalized recurrence period density entropy and the formula is given. But there is no such reference mentioned in the julia RQA package reference documentation. This leads to a series of puzzling questions. As mentioned above, I don't know whether the "recurrence time entropy" mentioned in the discussion above the proposal refers to the normalized recurrence period density entropy (RPDE), if the "rt_entropy" in the software refers to the normalized recurrence period density entropy , it is suggested that you can clearly indicate the original name of the indicator as "normalized recurrence period density entropy" in the description of "rt_entropy" in the reference document of the package, and give the formula, which is very important, because there is no part in the original paper 《Exploiting Nonlinear Recurrence and Fractal Scaling Properties for Voice Disorder Detection》 Call it "recurrence time entropy" or "RTE", which avoids a lot of confusing ambiguity.
Thank you so much for your tremendous effort and admirable work.

pagesys:
Closes #146 

Add the original nomenclature and references of the indicator to the rt_entropy function to improve the documentation notes

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RecurrenceAnalysis"
 uuid = "639c3291-70d9-5ea2-8c5b-839eba1ee399"
 repo = "https://github.com/JuliaDynamics/RecurrenceAnalysis.jl.git"
-version = "2.0.3"
+version = "2.0.4"
 
 [deps]
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
--- a/src/rqa/rqa.jl
+++ b/src/rqa/rqa.jl
@@ -114,14 +114,33 @@ macro histogram_params(keyword, description, hist_fun)
         fname = Symbol("$(keyword)_$(name)")
         _fname = Symbol("_$(keyword)_$(name)")
         fbody = function_bodies[name]
-        doc = """
-                $fname(R[; lmin=2, theiler])
+        if name == :entropy
+            doc = """
+                    $fname(R[; lmin=2, theiler])
 
-            Calculate the $(param) contained in the recurrence matrix `R`,
-            ruling out the lines shorter than `lmin` (2 by default) and all the
-            points inside the Theiler window (see [`rqa`](@ref) for the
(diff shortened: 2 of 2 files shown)


Datseris:
Sorry for the late response, thanks for the addition. Please also go to Project.toml and increment patch version by 1 (third number), so that I can immediatelly tag a new release with your change.

pagesys:
> Sorry for the late response, thanks for the addition. Please also go to Project.toml and increment patch version by 1 (third number), so that I can immediatelly tag a new release with your change.

It doesn't matter. I have already committed again after following your suggestion to modify version in the Project.toml file to 2.0.4, please check.

PR #150, opened 2023-03-02
Fix broken argument in windowed function

allomulder:
Hopefully this one fixes both the function and the associated docs.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RecurrenceAnalysis"
 uuid = "639c3291-70d9-5ea2-8c5b-839eba1ee399"
 repo = "https://github.com/JuliaDynamics/RecurrenceAnalysis.jl.git"
-version = "2.0.4"
+version = "2.0.5"
 
 [deps]
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
--- a/src/rqa/histograms.jl
+++ b/src/rqa/histograms.jl
@@ -107,8 +107,8 @@ end
 
 deftheiler(x::Union{RecurrenceMatrix,JointRecurrenceMatrix}) = 1
 deftheiler(x::CrossRecurrenceMatrix) = 0
-deftheiler(x::SparseMatrixCSC) = 1
-function diagonalhistogram(x::Union{ARM,SparseMatrixCSC}; lmin::Integer=2, theiler::Integer=deftheiler(x), kwargs...)
+deftheiler(x::AbstractMatrix) = 1
+function diagonalhistogram(x::Union{ARM,AbstractMatrix}; lmin::Integer=2, theiler::Integer=deftheiler(x), kwargs...)
     (theiler < 0) && throw(ErrorException(
         "Theiler window length must be greater than or equal to 0"))
     (lmin < 1) && throw(ErrorException("lmin must be 1 or greater"))
@@ -153,7 +153,7 @@ function diagonalhistogram(x::Union{ARM,SparseMatrixCSC}; lmin::Integer=2, theil
 end
 
 
-function verticalhistograms(x::Union{ARM,SparseMatrixCSC};
+function verticalhistograms(x::Union{ARM,AbstractMatrix};
     lmin::Integer=2, theiler::Integer=deftheiler(x), distances=true, kwargs...)
     (theiler < 0) && throw(ErrorException(
         "Theiler window length must be greater than or equal to 0"))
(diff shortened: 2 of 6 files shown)


Datseris:
Thank you! Alright, so here are the steps:

- This error wasn't caught because there is no test for it. Could you write a test that calls this windowed function in the test suite?
- Please incremet _patch_ version (3rd number) in Project.toml
- Please **revert** the formatting changes you did. You changed a lot of spacing but there is no reason for this. In the future we will have an automatic formatter.

allomulder:
@Datseris 

1. Never really wrote unit tests before, so I guess it's time to learn. So I simply use the `@testset` macro to create a set of windowed function tests and then just `@test some_condition`, do I get the logic right? 

Also I see that there are several .jl files inside the test folder. Should I create a separate file for this occasion or should I make it a part of `rmatrix_analytic`?  It seems like the second is easier but perhaps it doesn't fit the logic of that particular test file since it's about thresholds.

2. Got it, will do.

3. That's VS Code Formatter :c Will turn it off and recover the indentation.

allomulder:
Also while I'm at it: shouldn't this be `width-1`? If I understand this correctly, right now it's one step longer than specified width - for example, if I specify `width=30`, the sequence `windows[i]:(windows[i]+width)` should have length 31. Is this an expected behaviour or should I fix that too?

image

Datseris:
Okay, this is an easy fix. Replace the call to `nnz` to the call to `count`, as they effectively do the same thing but `count` is more general. This may come with a slight performance cost however. Could you please measure the performance cost? do something like
```julia
using BenchmarkTools
R = random recurrence matrix
@btime recurrencerate(R)
```
and then do exactly the same after changing `nnz` to `count`.

allomulder:
@Datseris 

Sorry, just got to the PC. Here is the comparison between old and version of the `recurrencerate`:

image

Used (basically) the same setup you have in your recurrence matrix threshold tests:

`t = range(0, 2π; length = 20)
c = cos.(t)
s = sin.(t)
X = StateSpaceSet(c, s)

dmat = distancematrix(X)
threshold = dmat[1, 2] + 0.01 

rec = RecurrenceMatrix(X, threshold)`

Here are the results of the original function benchmark:

image

And here are the results of the second one:

image

Overall the performance cost seems a bit nasty, although better than I thought it would be.

allomulder:
I've also tested the alternative version of `windowed` function - without using views:

image

Here is the benchmark of the original `windowed`, but new `recurrencerate`:

image

And here is the benchmark of the new `windowed` and old `recurrencerate`:

image

Overall the new `windowed` function is faster, but comes with the higher memory cost (I guess that is the reason views were used in the first place).

Datseris:
I think we should change the `windowed` function to _not_ use views. It seems to me that views do not work well with sparse matrices. The base Julia langauge does not have the optimizations we would want there.

SO, if we change the windowed version to the allocating one, all functions work fine without modification, right? 

(nevertheless, we still should keep the modification that changes to using `AbstractMatrix` in the method decleration)

Datseris:
in the recurrence rate you can add an if block: 

```
if R isa Union{ARM, SparseMatrix}
nnz(R)
else
count(R)
end
```
so that it can still work in abstract matrices if someone gives one in the future.

allomulder:
@Datseris 

Okay, that seems to be working! Now for my next suggestions.

Let's assume I want to calculate recurrence rate over 30-point nonoverlapping windows. Let's also assume I have a 300-point signal:

image

What seems logical to me:

* `width` and `step` parameters should be equal to 30 each
* I should get a 10-point Vector as a result (300 / 30 = 10 epochs)

In its current state this function returns a **9-point Vector**. I did a slight modification that works correctly:

 
image

The modification consists simply of subtracting and adding 1 where appropriate:

image

Would it be okay if I add this too?

allomulder:
In the same vein:

1. Perhaps we should add an assertion that `width` cannot be less than 2 (since 1-point window is just the same value, and anything less just doesn't make sense). `step` should also be at least 1 - I can't really think of use for negative and 0-length steps.

2. Argument types should also probably be declared - `rmat` as ARM/AbstractMatrix, `f` as Function, and `width` + `step` as Integers.

Datseris (approved):
Fantastic, thanks for the contribution. Please go into Project.toml and increment the patch version number (third number) so we can release this as a bugfix release!

Datseris:
for your point 1: yes. For your point 2: the only benefit in doing that is that it increases code verbosity, which may also increase code clarity. but the actual code stays unafected witht he extra declretaions. Feel free to add them!

allomulder:
@Datseris 

Done all of that! Also added some tests as you asked - they test if all RQA functions work with `windowed` and if they return a vector of expected length. Perhaps they could be modified later. but for now they seem to work okay.

allomulder:
@Datseris 

Hello! Just a reminder, in case you've missed this :)

Datseris:
Thank you very much, i did forget, sorry!

PR #170, opened 2025-04-10
Fix the threading model to use user-defined tasks instead

asinghvi17:
It makes the code a bit more annoying to read through, but at least it should be correct now...


asinghvi17:
Somehow, the threaded one is taking 5 seconds and the non-threaded one is 100x less...I must be doing something wrong, but not sure what!

PR #178, opened 2026-02-02
Fix Dict method declaration for RQA type

KristianHolme:
Fixes this warning:
```
  1 dependency had output during precompilation:
┌ RecurrenceAnalysis
│  WARNING: Constructor for type "Dict" was extended in `RecurrenceAnalysis` without explicit qualification or import.
│    NOTE: Assumed "Dict" refers to `Base.Dict`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Dict end`.
│    Hint: To silence the warning, qualify `Dict` as `Base.Dict` in the method signature or explicitly `import Base: Dict`.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "RecurrenceAnalysis"
 uuid = "639c3291-70d9-5ea2-8c5b-839eba1ee399"
 repo = "https://github.com/JuliaDynamics/RecurrenceAnalysis.jl.git"
-version = "2.1.3"
+version = "2.1.4"
 
 [deps]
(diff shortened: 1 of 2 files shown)


Datseris (approved):
{Please als ogo to Project.toml and increase patch version by one!

KristianHolme:
> {Please als ogo to Project.toml and increase patch version by one!

Done!

Datseris (approved):

